PR #2, opened 2017-07-25
Add power DARN instruction support

nhorman:
Detect power arch build and enable support for DARN random number instruction.
If enabled, at runtime, detect available instruction support, and, if present,
use DARN to seed an AES encryption operation of a garbage buffer to generate
random data for use in adding entropy to kernel rng.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,6 +16,10 @@ if RDRAND
 rngd_SOURCES	+= rngd_rdrand.c rdrand_asm.S
 endif
 
+if DARN
+rngd_SOURCES	+= rngd_darn.c
+endif
+
 rngd_LDADD	= librngd.a
 rngd_LDFLAGS	= -lsysfs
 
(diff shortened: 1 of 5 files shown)


grooverdan:
I've measured DARN to output ~70-80MB/s so why use AES?

On 4GiB of test data:
<pre>
$  ent ~/darn_data.dd
Entropy = 8.000000 bits per byte.Optimum compression would reduce the size of this 4294967296 byte file by 0 percent.
Chi square distribution for 4294967296 samples is 226.07, and randomly would exceed this value 90.38 percent of the times.
Arithmetic mean value of data bytes is 127.5012 (127.5 = random).
Monte Carlo value for Pi is 3.141484506 (error 0.00 percent).
Serial correlation coefficient is 0.000004 (totally uncorrelated = 0.0).
</pre>

also FYI - kernel support directly is on way: https://patchwork.ozlabs.org/patch/785381/

PR #4, opened 2017-07-28
parse_opt: use right type in printf

grooverdan:
Previous error:

rngd.c: In function ‘parse_opt’:
rngd.c:195:44: warning: format ‘%d’ expects argument of type ‘int’, but
argument 2 has type ‘long int’ [-Wformat=]
    printf("exclude index is out of range: %d\n", idx);
                                            ^
rngd.c:199:22: warning: format ‘%d’ expects argument of type ‘int’, but
argument 2 has type ‘long int’ [-Wformat=]
   printf("Disabling %d: %s\n", idx, entropy_sources[idx].rng_name);
                      ^


grooverdan:
fixed by https://github.com/nhorman/rng-tools/commit/4facbc23d62abcfadd272dd784476a939f89d567

PR #6, opened 2017-07-28
Check darn hardware with getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_3_00

grooverdan:
Simpler and more correct.

Tested:

$ ./rngd --list
Failed to init entropy source 0: Hardware RNG Device

Failed to init entropy source 1: TPM RNG Device

Enabling power DARN rng support

Available entropy sources:
3: Power9 DARN Instruction RNG

### DIFF
--- a/rngd_darn.c
+++ b/rngd_darn.c
@@ -28,18 +28,9 @@
 #error power DARN support requires libgcrypt!
 #endif
 
-#include <unistd.h>
-#include <stdint.h>
 #include <stdlib.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
-#include <errno.h>
-#include <syslog.h>
-#include <string.h>
-#include <stddef.h>
 #include <limits.h>
-#include <sysfs/libsysfs.h>
+#include <sys/auxv.h>
 #include <gcrypt.h>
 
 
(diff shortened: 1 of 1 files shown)


nhorman:
you shouldn't need to test for the availability of AT_HWCAP or PPC_FEATURE2_ARCH_3_00, as the former should always be defined on linux and the latter should always be defined on linux/power, so you can remove the #if/#else statements and reduce the code.  Also, unless I'm mistaken, While  PPC_FEATURE2_ARCH_3_00 implies DARN support, so does PPC_FEATURE2_DARN no?  I would think that using the latter rather than the former would make the code a bit more clear.

grooverdan:
Didn't know about PPC_FEATURE2_DARN - quite recent (https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg168253.html ) and make sense. On the assumption that glibc/kernel distro backports happen, sure, lets remove the #if/else. Updated PR coming soon.

grooverdan:
Updated (and removed a few no-longer needed includes/definitions).

I'm assuming AES is because you are going off the architected promise of randomness which is less than delivered (measured results https://github.com/nhorman/rng-tools/pull/2#issuecomment-318942418 )

nhorman:
In answer to your question, yes, The system I measured on provided far less entropy than your measurements.  Specifically, multiple calls on the instruction in close succession returned values that were both not ULONG_MAX (implying a successful call), and identical to the previous call to the instruction.  So it acted as though it provided high throughput, but didn't seem safe to me as an RNG entropy source by itself.  As a result I decided to use AES to generate pseudo ramdom data with the output of DARN as a key (with periodic rekeying of course).  This is similar to the method of entropy generation that the intel RDRAND instruction uses.

PR #16, opened 2018-07-04
Addition of CPU Jitter RNG noise source

smuellerDD:
The CPU Jitter RNG noise source is added as a new generic noise source.
It requires the static CPU Jitter RNG library that can be obtained from
[1]. The static library can be pointed to using the configure option of
--with-libjitterentropy-prefix.

The CPU Jitter RNG noise source is extensively discussed at [1].

[1] http://www.chronox.de/jent.html

Signed-off-by: Stephan Mueller <smueller@chronox.de>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -11,11 +11,18 @@ noinst_LIBRARIES = librngd.a
 
 rngd_SOURCES	= rngd.h rngd.c	rngd_entsource.h rngd_entsource.c	\
 		  rngd_linux.h rngd_linux.c util.c
+rngd_LDADD	=
+rngd_CFLAGS	=
 
 if NISTBEACON
 rngd_SOURCES	+= rngd_nistbeacon.c
 endif
  
+if JENT
+rngd_SOURCES	+= rngd_jent.c libjitterentropy.a
+rngd_LDADD	+= $(libjitterentropy_prefix)/libjitterentropy.a
+endif
+
 if RDRAND
 rngd_SOURCES	+= rngd_rdrand.c rdrand_asm.S
 endif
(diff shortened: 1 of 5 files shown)


nhorman:
So, I've got my branch that implements this as well (see the jitterentropy branch).  It looks like we both have very similar code,but I'd like us to look at both before we decide which one to pull

smuellerDD:
Am Donnerstag, 5. Juli 2018, 02:46:34 CEST schrieb Neil Horman:

Hi Neil,

> So, I've got my branch that implements this as well (see the jitterentropy
> branch).  It looks like we both have very similar code,but I'd like us to
> look at both before we decide which one to pull

Just looked at your code:

- the while loop in xread_jitter would not be needed

- shouldn't *ec be static?

- very nice that there is a close function, I was not aware that there is such 
close operation.


Ciao
Stephan

nhorman:
Thank you, the close routine was added specifically to tear down the entropy collector instance here, but I plan to use it to clean up the nist entropy beacon entropy source as well.

I've cleaned up your other notes and created an alternate pull request here:
https://github.com/nhorman/rng-tools/pull/17

The question then is, which one should we use?  Yours looks good as well, but could use a few syntatical clean ups.  Or should we merge them into a single pull request?

smuellerDD:
> Am 05.07.2018 um 14:02 schrieb Neil Horman <notifications@github.com>:
> 
> The question then is, which one should we use? Yours looks good as well, but could use a few syntatical clean ups. Or should we merge them into a single pull request?
> 

With the updates I think your patch should be used. The rngd_jitter.c is almost identical to mine. The difference is in the autoconf and library integration which you have solved nicely.

Reviewed-by: Stephan Mueller <smueller@chronox.de>

nhorman:
Ack, then I'll say thank you for all your help, close this pull request, merge mine and get it integrated to fedora and RHEL.

Greatly appreciate the help here!

smuellerDD:
Just saw one error: xread_jitter must use int ret instead size_t.

Cui bono?

PR #18, opened 2018-07-05
Catch negative error codes in xread_jitter

smuellerDD:
The jent_read_entropy function returns negative error codes. Thus, the
ret variable must be signed.

Signed-off-by: Stephan Mueller <smueller@chronox.de>


nhorman:
Thanks!

PR #19, opened 2018-07-08
rngd.c: fixed main loop logic, and a couple of off-by-one errors

lonwillett:
This fixes a critical error in "do_loop" which prevented entropy source 0 (/dev/hwrng) from ever being used.

It also fixes a couple of off-by-one errors, and makes do_loop function more like it was intended (I think).

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -308,13 +308,12 @@ static int update_kernel_random(int random_step,
 static void do_loop(int random_step)
 {
 	unsigned char buf[FIPS_RNG_BUFFER_SIZE];
-	int retval = 0;
-	int no_work = 0;
-	int i = 0;
+	int no_work;
+	bool work_done;
 
-	while (no_work < 100) {
+	for (no_work = 0; no_work < 100; no_work = (work_done ? 0 : no_work+1)) {
 		struct rng *iter;
-		bool work_done;
+		int i, retval;
 
 		work_done = false;
 		for (i = 0; i < ENT_MAX; ++i)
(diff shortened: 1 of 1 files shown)


nhorman:
This looks pretty good, but  has a few issues:

* The for loop is wrong, but I would like to keep the indexing logic there, rather than inside the loop.  Please modify it so that the for loop looks like this:
for(;;i=(++i % ENT_MAX))

That should iterate you from 0 to ENT_MAX-1 for ever

* I'm not sure why your resetting the no_work value in an else clause.  I think you should just reset it to zero right before breaking out of the for  loop, no?

lonwillett:
I'm not sure I've understood all the intended logic, but it looks to me like one wants to exit the for-loop after every pass through the entropy sources, so as to increment and check no_work. So "i" really does need to hold ENT_MAX+1 values, to distinguish the states:

- Get entropy from source 0
- Get entropy from source 1
...

- Get entropy from source ENT_MAX-1
- Check that work was done, and reset the value to start over

The current code is buggy because it effectively removes the "get entropy from source 0" state.

Your proposed change would have a less serious bug, in that it would remove the "check that work was done" state; if all the entropy sources get disabled, then the for-loop would never exit (though it would start burning CPU time doing nothing).

Reading through the code more closely, it appears to me that the early exit from the for loop (the "break" when an entropy source is successfully used) is actually pointless; the value of work_done will have been changed to reflect this success. So it probably makes more sense to just remove this, and change the loop to a bog-standard for loop. I'll submit that. It will probably be clearer. Tell me if there is something that I am failing to understand.

nhorman:
Yes, this actually looks much better, thank you!

PR #30, opened 2018-09-06
Add cross-compile fixes

abelbeck:
Check for 'ar' as it is used for standard builds as well as jitterentropy
Export select variables when building the jitterentropy library


nhorman:
Acked-by: Neil Horman <nhorman@tuxdriver.com>

PR #33, opened 2018-09-17
Jitter: adjust buffer refill for use while gathering entropy

abelbeck:
Revert edecf86 since it causes a 5x performance penalty built --without-libgcrypt
As such, only generate jent_read_entropy for what is needed (buf_sz - avail)
If 'avail' changes while jent_read_entropy() is called, account for it by
idx = pre-gather-avail - post-gather-avail
As such, only overwrite used entropy and maintain unused entropy
Note: while testing (pre-gather-avail - post-gather-avail) was always 0, not without trying to make them different
so this may not be a common problem, but this accounts for it should it ever occur.

A ring-buffer solution would be an alternate solution, but looks like it may be over-kill.

### DIFF
--- a/rngd_jitter.c
+++ b/rngd_jitter.c
@@ -264,6 +264,7 @@ static void *thread_entropy_task(void *data)
 	cpu_set_t cpuset;
 
 	ssize_t ret;
+	size_t need;
 	struct thread_data *me = data;
 	char *tmpbuf;
 	struct timespec start, end;
@@ -321,9 +322,10 @@ static void *thread_entropy_task(void *data)
 			break;
 
 		/* We are awake because we need to refil the buffer */
+		need = me->buf_sz - me->avail;
 		pthread_mutex_unlock(&me->mtx);
 		clock_gettime(CLOCK_REALTIME, &start);
-		ret = jent_read_entropy(me->ec, tmpbuf, me->buf_sz);	
+		ret = jent_read_entropy(me->ec, tmpbuf, need);
 		clock_gettime(CLOCK_REALTIME, &end);
 		message(LOG_DEBUG|LOG_ERR, "jent_read_entropy time on cpu %d is %.12e sec\n",
 			me->core_id, elapsed_time(&start, &end));
@@ -334,9 +336,10 @@ static void *thread_entropy_task(void *data)
 		update_sleep_time(me, &start, &end);
 		if (!me->buf_ptr) /* buf_ptr may have been removed while gathering entropy */
 			break;
-		memcpy(me->buf_ptr, tmpbuf, me->buf_sz);
-		me->idx = 0;
-		me->avail = me->buf_sz;
+		/*  idx = pre-gather-avail - post-gather-avail */
+		me->idx = ((me->buf_sz - me->avail - need) > 0) ? (me->buf_sz - me->avail - need) : 0;
+		memcpy(me->buf_ptr + me->idx, tmpbuf, need);
+		me->avail = me->buf_sz - me->idx;
 		me->refill = 0;
 
 	} while (me->buf_ptr);


nhorman:
As I explained in our previous conversation, while I understand this returns performance to a level that you like, that doesn't mean it works.  With this revert, we use idx as both a buffer read and fill index, which is bad because we don't keep the per-thread mutex lock held while generating new entropy.  That means that, while running jent_read_entropy, the main thread can come in, read some more entropy from the buffer, and advance idx.  If that happens, when the per-cpu thread comes back and writes in the new entropy, it will do so using the old computed value of need, and the new value of idx, leaving some old used entropy in place, and potentially overwriting unused entropy if the buffer wraps.  We need to add a separate fill index that is only incremented on a refill operation, which is effectively a ring buffer approach.  I've started working on that but got sidetracked today.  If you would like to take a stab at it, please feel free.  All we should need is:
1) another per-thread variable fill_idx, that tracks the start of the used random data
2) A generation of need bytes of random data on wake up (which your patch above has)
3) A copy of the data in (2) to the per thread buffer starting at fill_idx, and clamped at buf_sz-buf_avail bytes
4) An alteration of the gcrypt code to do a similar random data and copy

abelbeck:
While a redesign may be easier to understand, this PR does take into account for a changing `idx` and copies the new entropy offset by the changed `idx` so all is good.

Also interesting is I have yet to see `idx` / `avail` change while calculating entropy, which I find surprising.

abelbeck:
Hi @nhorman I have re-run the logic of this PR in my head, imagine edge conditions, and I can't think of case where this PR doesn't do what we want.  And it is a simple change.

Do you agree ?  If not, where does it fail ?

nhorman:
I do not agree.  consider this case:

start conditions: rngd has read 6 bytes of data from a 10 byte buffer, meaning bytes 0-5 (inclusive) have been used, and need refilling.  Start conditions for the buffer are:
idx = 6
buf_sz = 10
avail = 4

operational flow:
|----------------reader----------------------|---------------writer------------------------|
|:---------------------------------------------:|:---------------------------------------------:|
|  reader cond signals writer---------  |  ---------------------------------------------|
|-----------------------------------------------|  writer wakes up and locks mtx-----|
|-----------------------------------------------| compute need=10-4=6---------------|
|-----------------------------------------------| unlock mutex----------------------------|
|------lock mutex read 2 bytes---------| -call jent_read_entropy(6 bytes)----|
|-----idx = 8, avail = 2 -------------------| -----------------------------------------------|
|-----unlock mtx---------------------------|----return from jent_read,lock mtx-----|
|----------------------------------------------|--------compute new idx-------------------|
|----------------------------------------------|--(see A below)-----------------------------|
|----------------------------------------------|  writer overwrites need bytes at idx 2|



(A) me->idx = ((me->buf_sz - me->avail - need) > 0) ? (me->buf_sz - me->avail - need) : 0;
      me->idx = (10-2-6) > 0 ? (10-2-6) : 0
      me->idx = 2 > 0 ? 2 : 0
      me->idx = 2

The generic form of the issue is that, during the generation phase where jent_read_entropy is called by the writer, the lock is release (correctly, as the reader should be able to come read out more entropy).  During that time, the avail and idx values are updated, and as a result the new idx value that you compute will be off by exactly the cumulative amount that avail changed during that period.  Above you can see that idx should have been set to zero because, as stated in the initial conditions bytes 0-5 (inclusive had already been read).  What this has done, is left 2 bytes at the start of the array un-refilled, while the two remaining bytes at the end of the array, which were not yet used get overwritten.

Your solution works only in the case where reads and writes are serialized and toggling (i.e. the writer can completely refill the buffer in a single operation.  We can achieve that by holding the mutex during the call to jent_read_entropy, but that will re-introduce the large latency and low throughput issues we had previously.

Really, the correct solution is a ring buffer in which we track both the read and the write point and update them modulo the buffer size.  Its not super complex and I'm about 2/3's done with it, and will post it as soon as I clear out my work backlog

abelbeck:
> (A) me->idx = ((me->buf_sz - me->avail - need) > 0) ? (me->buf_sz - me->avail - need) : 0;
me->idx = (10-2-6) > 0 ? (10-2-6) : 0
me->idx = 2 > 0 ? 2 : 0
me->idx = 2

Exactly, and the refill adjusts accordingly
```
		memcpy(me->buf_ptr + me->idx, tmpbuf, need);
		me->avail = me->buf_sz - me->idx;
```
The 6 `need` bytes get placed at 2-7 and `idx` gets rewound from 8 to 2.

Bytes 8-9 are left alone as they are unused.

abelbeck:
I added a fix for (1), I'm quite surprised that I can't trigger a non-zero `idx` on my 1-GHz 4-core CPU, though I artificially created a `idx` of 3 and 5 to test the code.

Note that I fixed a `ret` test error in this code's neighborhood.

I think I will pass on (2) as you understand that much better than I.

nhorman:
Its a bit wonky, I agree.  I think getting a non zero idx is going to be an artifact of the rate at which you drain /dev/random.  cat /dev/random > /dev/null will drain the entropy pool much much faster than jitter can fill it, so its likely just toggling between completely full and completely empty.  I wonder if its worth adding a entropy drain utility that we can use to finely tune the rate at which we drain an entropy source

regarding your new patch, I was really hoping that, instead of doing a second check and fill, you would just check need again, and, if it was non-zero, execute a goto back into the loop that would reuse the existing code to preform another fill operation.  That way if we were draining constantly the thread would never go back to sleep.  Can you do it that way? It would be less code, and more efficient.

abelbeck:
Testing on my slow 0.5 GHz 1-core box, same results, no `on cpu %d filled in missing` logs.

But I do see a fair number of `JITTER skips empty thread on cpu %d` logs. (Also on my 4-core box)

I hope a light bulb just went off in your head :-)

Could it be that when `refill` is true any additional request will be directed to the next thread, as such a non-zero `idx` is never true ?

nhorman:
of course you can re-use the majority of the code, it does exactly what you need it to, just enclose everything between the computation of need and the setting of me->avail, enclose it in a do { }while(need) loop, and recompute the value of need at the very end of the loop (with an optional message conditional on !need indicating that we have more entropy to generate. It should be about 5 lines of code.

As for the non zero idx value, I just found that problem, xread_jitter uses the ->refill flag to determine if it skips a buffer, but ->refill only means that the buffer is in the buffer is generating new entropy (not that the buffer is empty).  xread_jitter should key instead off of ->avail being equal to zero to determine buffer emptyness.  In fact the entire ->refill flag can be removed

abelbeck:
Thanks, I'll work on removing `->refill`.

nhorman:
Yeah, originally we had to do things in powers of two, but now any size
drain or fill should be fine

Neil

nhorman:
thank you, the refill removal and avail check looks good, I appreciate your commitment to making this right.  I still need you to fix that middle commit though please, I really don't just want a single check after the first fill.  I'd like the writer threads to continuously refill until such time as their buffers have no more empty space without needing to go through a pthread_cond_wait cycle.

abelbeck:
OK, I'll work on continual processing in the thread_entropy_task() `do { } while (me->buf_ptr)` loop.

nhorman:
thank you

abelbeck:
Take a look at 9cc82de , I think it covers your concerns.

On my 1-GHz 4-core box, it is now ~60% faster in creating entropy, at ~15 Kbps.

Seems to work nicely in testing, creating a dynamic balance between filling entropy and draining entropy.

The `pthread_mutex_trylock()` addition to `xread_jitter()` solved a critical problem created by removing the `refill` variable in the previous commit here.

BTW, there is an ultra-small chance a `pthread_mutex_lock()` in the thread_entropy_task() would occur while `xread_jitter()` had the lock. Is this an issue ? if so what is the standard solution ... a `pthread_mutex_trylock()` loop of `usleep()` until trylock returns true ?

nhorman:
That looks pretty good. Thanks, and nice performance numbers!  One nit, the
message regarding filling of the buffer, should be LOG_DEBUG or LOG_INFO,
not LOG_ERR, but I'll fix that after I pull the branch tomorrow, so don't
worry about it.  I also would have looped need not being zero rather than
idx > 0, bit yours is probably.more in keeping with your fill computation
anyway.

Regarding the use of pthread_lock, it should be fine, if the fill task
holds the lock the read task will just block until its released, which
should be a small window since we release the lock quickly in the writer.
I will say that your reversion to theuse of trylock might be detrimental,
because a locked take just implies that a thread is modifying the buffer or
it's metadata, not that the buffer is empty (i.e. it's possible that if you
just wait for the lock, you'll.find fresh entropy there).  But as long as
you have more than one core, it likely won't matter, as you aren't likely
to hit 4 locks.in a row.  A good test would.be to compare the performance
of trylock vs lock on a.single could system that only starts a single write
thread.

abelbeck:
> One nit, the message regarding filling of the buffer, should be LOG_DEBUG or LOG_INFO, not LOG_ERR

Yes I noticed that as well, but I just kept it constant to the other occurrences.

BTW, without trylock in xread_jitter() `idx` was being incremented while the thread was locked in thread_entropy_task().  Changing lock to trylock fixed that, previously testing for `refill` kept that from being a problem.  Fortunately a locked thread occurs less often than the old `refill` flag so that is a general performance improvement.

I did test on a 1-core and it seemed to work fine.  The key is entropy generation now toggles between locked and unlocked so both filling and draining gets time when needed.

nhorman:
Wait, I'm not sure I understand.  When you say "without trylock", do you
actually mean "with trylock instead of a regular mutex lock"?   Or do you
mean "withoit any locks at all"?  If you mean the latter, then that makes
sense, as you need locking on both the read and write side to prevent
mutual exclusion, but you should be able to use with trylock or just plain
lock on the read side, you just need to check the return code in the former
case.

If performance.is good on a single core though, then the trylock/lock
question is pretty well moot in my mind.  We can experiment later if
there's value, but it's good right now as is.

Neil

abelbeck:
> Wait, I'm not sure I understand.  When you say "without trylock", do you
actually mean "with trylock instead of a regular mutex lock"? 

Yes, the change here on Line 156
https://github.com/nhorman/rng-tools/pull/33/commits/9cc82de5f40f6f529bcc05bab1cbef859e8a9eb4#diff-125de59086175be28f557034dddaec06L156

Without that change, in xread_jitter() `idx` was being incremented while the thread was locked in thread_entropy_task().  I swear :-)  Again, previously, testing for `refill` kept that from being a problem.

nhorman:
I really can't believe that, but let's get this check in in the morning and
worry about that later.  Either way the locking should.be correct

abelbeck:
With more testing on faster CPU's I'm still seeing `idx` changing on a locked thread, turns out we have a superfluous `pthread_mutex_unlock` in xread_jitter().  This fixes the problem
```
--- rng-tools-6.4/rngd_jitter.c.orig	2018-09-20 10:55:07.570289235 -0500
+++ rng-tools-6.4/rngd_jitter.c	2018-09-20 11:25:35.009033322 -0500
@@ -214,12 +214,8 @@
 		}
 	}
 	rc = 0;
-
-	pthread_mutex_unlock(&current->mtx);
-	rc = 0;
 out:
 	return rc;
-
 }
 
 static inline double elapsed_time(struct timespec *start, struct timespec *end)
```
This might also explain why `pthread_mutex_trylock()` helped, but was not the source of the problem.  Though I think `pthread_mutex_trylock()` and trying another core is better than spinning waiting for the thread to unlock.

Also the `sleep` structure may not always be initialized with the removal of `refill`.

I'll work up another PR for these.

PR #34, opened 2018-09-20
Jitter: remove superfluous pthread_mutex_unlock

abelbeck:
Also initialize the sleep timespec to 1 second in case it is not dynamically initialized

### DIFF
--- a/rngd_jitter.c
+++ b/rngd_jitter.c
@@ -147,7 +147,7 @@ int xread_jitter(void *buf, size_t size, struct rng *ent_src)
 	char *bptr = buf;
 	int rc = 1;
 	int retry_count = 0;
-	struct timespec sleep;
+	struct timespec sleep = { 1, 0 };
 try_again:
 	while (need) {
 		/* if the current thread is locked or an empty buffer
@@ -214,12 +214,8 @@ next:
 		}
 	}
 	rc = 0;
-
-	pthread_mutex_unlock(&current->mtx);
-	rc = 0;
 out:
 	return rc;
-
 }
 
 static inline double elapsed_time(struct timespec *start, struct timespec *end)


nhorman:
I agree on the superfluous unlock that got introduced, but:

1) please use  separate commits for the sleep and mutex fixes, so we have the documented clearly.

2) please don't just init the sleep value to 1 second, thats not he documented behavior.  It should always be initialized prior to use, but its value should be taken from the slptime variable always.  Just move the memcpy from slptime either to right before the call to nanosleep, or to the top of the function (preferably the latter, so we don't incur the overhead of the copy unless we need it)

abelbeck:
> please don't just init the sleep value to 1 second, thats not he documented behavior. It should always be initialized prior to use, but its value should be taken from the slptime variable always. Just move the memcpy from slptime either to right before the call to nanosleep, or to the top of the function (preferably the latter, so we don't incur the overhead of the copy unless we need it)

If we initialize `sleep` at the top of the function (before `try_again:`), we could add:
```
pthread_mutex_lock(&current->mtx);
memcpy(&sleep, &current->slptm, sizeof(struct timespec));
pthread_mutex_unlock(&current->mtx);
```
But that seems costly to possibly spin to just get `sleep`.

Is that what you want ?

A less costly, but less pure approach...
```
int sleep_needs_init = 1;
struct timespec sleep = { 1, 0 };

... when we have a lock ...
if (sleep_needs_init) {
  memcpy(&sleep, &current->slptm, sizeof(struct timespec));
  sleep_needs_init = 0;
}

So if a lock is never reached the default 1 second would be used, but unlikely on multi-core CPU's.
```
Thoughts ?

nhorman:
I've already fixed this.

PR #36, opened 2018-09-22
Test Mode: various enhancements and tweaks

abelbeck:
Keep track of elapsed time, don't count on alarm(1) for timing
Use first alarm(1) as t=0
Use bits instead of bytes
Various output text tweaks

Sample results:
Entropy rate:  5.408 Kbits/sec averaged over 18 iterations for  65.01 seconds
Entropy rate:  52.34 Kbits/sec averaged over 190 iterations for    275 seconds
Entropy rate:  2.678 Mbits/sec averaged over 15 iterations for  15.06 seconds
Entropy rate:  104.2 Mbits/sec averaged over 26 iterations for     26 seconds
Entropy rate:    126 Mbits/sec averaged over 26 iterations for     26 seconds

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -48,6 +48,7 @@
 #include <signal.h>
 #include <limits.h>
 #include <ctype.h>
+#include <time.h>
 
 #include "rngd.h"
 #include "fips.h"
@@ -138,7 +139,8 @@ struct arguments *arguments = &default_arguments;
 
 static unsigned long ent_gathered = 0;
 static unsigned long test_iterations = 0;
-static double avg_entropy = 0;
+static double sum_entropy;
+static struct timespec start_test, end_test;
 
 static enum {
 	ENT_HWRNG = 0,
@@ -498,7 +500,7 @@ static int random_test_sink(struct rng *rng, int random_step,
 {
 	if (!ent_gathered)
 		alarm(1);
-	ent_gathered += FIPS_RNG_BUFFER_SIZE;
+	ent_gathered += (FIPS_RNG_BUFFER_SIZE * 8);
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


abelbeck:
1) `sum_entropy` is a double, no overflow fears.  Possibly the `unsigned long` `ent_gathered`, though reset every 1-ish seconds is 32-bits on 32-bit systems which still has a 30x margin (per my testing) using the fastest (rdrand), but I can change that to bytes to give us 240x margin on 32-bit systems if you want.

2) I think this time calculation is correct...
```
test_time = (end_test.tv_sec - start_test.tv_sec);
test_time = ((test_time * 1.0e9) + (end_test.tv_nsec - start_test.tv_nsec)) / 1.0e9;
```
For sake of illustration, let's use msec instead of nsec
end = { 3 , 100 }   ie. 3.1 seconds
start = { 1 , 800 }   ie. 1.8 seconds

test_time = ( 3 - 1 ) = 2
test_time = ((test_time * 1000) + (100 - 800)) / 1000 = (2000 - 700)/1000 = 1.3

Though your rngd_jitter `elapsed_time()` is not correct (I started with that :-)) and could be replaced with the above.

3) OK, I'll add defines

4) OK if you like that better, I prefer seeing it set when the clock is started for clarity, but no functional difference.

5) OK good point, that bothered me as well.

6) You did not mention it, but we should use CLOCK_MONOTONIC here.

nhorman:
you're right on the timspec calculation, I was thinking tv_nsec was unsigned which would lead to an improper result

abelbeck:
@nhorman Your changes have been implemented

nhorman:
Ack, thank you!

abelbeck:
Thanks for your review.

BTW, in `rngd.c` you should make `server_running` `volatile` as it is set by the signal handler.
```
- bool server_running = true;
+ volatile bool server_running = true;
```
Are we close to a 6.4.1 or 6.5 release ?

nhorman:
You're welcome

And we're not.changing the server_running to a volitle, that's what that
keyword is for.  Volatile forces the compilier to do exectly what you
wrote, rather than make optimizations that could have adverse effects on
intended operation (i.e. hardware access):


https://stackoverflow.com/questions/246127/why-is-volatile-needed-in-c

As for.the release, we can do it pretty well, anytime, i just want to make
a change to allow dynamic linking of jitterentropy, I'll.have it ready soon.

abelbeck:
Hi @nhorman , I'm confused ... it is my impression, (and per the stackoverflow link) that `volatile` tells the compiler to not optimize a `volatile` variable since it may be changed via external means, such as direct hardware access or a signal handler.  In ways the optimizer could not anticipate.

Note that `rngtest.c` uses a `volatile` for it's signal handler global...
https://github.com/nhorman/rng-tools/blob/master/rngtest.c#L176

A few months back I encountered such a case in `asterisk` (PBX) when a global pointer was being optimized out of the code, after much hair-pulling it became clear that the global (changed by a sub-process) required `volatile` and that fixed the problem.

I'm interested why you don't think `volatile` applies to the `server_running` global.

PR #37, opened 2018-09-25
Remove static

nhorman:
This is a pull request to remove the jitterentropy submodule from rng-tools.  The intent is to remove it and modify the configure script such that it search for the jitterentropy dso on the host system instead.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "jitterentropy-library"]
-	path = jitterentropy-library
-	url = https://github.com/smuellerDD/jitterentropy-library.git
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,18 +1,7 @@
 ##
 ## Toplevel Makefile.am for rng-tools
 ##
-if JITTER
- JSUBDIR = jitterentropy-library
- JSUBLIB = ./jitterentropy-library/libjitterentropy.a
- AM_CPPFLAGS = -I./jitterentropy-library
- export CC AR CFLAGS LDFLAGS
-endif
-
-if JITTER_DSO
- JSUBLIB= -ljitterentropy
-endif
-
-SUBDIRS		= contrib tests $(JSUBDIR) 
+SUBDIRS		= contrib tests 
 
 sbin_PROGRAMS	 = rngd
 bin_PROGRAMS	 = rngtest
@@ -38,12 +27,8 @@ if JITTER
 rngd_SOURCES	+= rngd_jitter.c
 endif
 
-if JITTER_DSO
-rngd_SOURCES	+= rngd_jitter.c
-endif
 
-
-rngd_LDADD	= librngd.a -lsysfs $(JSUBLIB) ${libcurl_LIBS} ${libxml2_LIBS} ${openssl_LIBS} $(PTHREAD_LIBS)
+rngd_LDADD	= librngd.a -lsysfs $(LIBS) ${libcurl_LIBS} ${libxml2_LIBS} ${openssl_LIBS} $(PTHREAD_LIBS)
 
 rngd_CFLAGS	= ${libxml2_CFLAGS} ${openssl_CFLAGS} $(PTHREAD_CFLAGS)
 rngd_LDFLAGS	= $(PTHREAD_CFLAGS)
--- a/configure.ac
+++ b/configure.ac
@@ -60,17 +60,11 @@ AM_CONDITIONAL([DARN], [test $target_cpu = powerpc64le])
 AS_IF([test $target_cpu = powerpc64le], [AC_DEFINE([HAVE_DARN],1,[Enable DARN])],[])
 
 AM_CONDITIONAL([JITTER], [false])
-AM_CONDITIONAL([JITTER_DSO], [false])
-AS_IF([test -f jitterentropy-library/Makefile],
+AC_SEARCH_LIBS(jent_version,jitterentropy,
 		[AM_CONDITIONAL([JITTER], [true])
-		 AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])],
-		[AC_SEARCH_LIBS(jent_version,jitterentropy,
-			[AM_CONDITIONAL([JITTER_DSO], [true])
-			 AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])],
-			[AC_MSG_NOTICE([Disabling JITTER entropy source])])])
-
-#AM_CONDITIONAL([JITTER], [test -f jitterentropy-library/Makefile])
-#AS_IF([test -f jitterentropy-library/Makefile], [AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])],[AC_MSG_NOTICE([Disabling JITTER entropy source])])
+		AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])
+		AC_MSG_NOTICE([Enabling JITTER entropy source])],
+		AC_MSG_NOTICE([Disabling JITTER entropy source]))
 
 AS_IF(
 	[ test "x$with_nistbeacon" != "xno"],
--- a/configure.ac
+++ b/configure.ac
@@ -60,11 +60,25 @@ AM_CONDITIONAL([DARN], [test $target_cpu = powerpc64le])
 AS_IF([test $target_cpu = powerpc64le], [AC_DEFINE([HAVE_DARN],1,[Enable DARN])],[])
 
 AM_CONDITIONAL([JITTER], [false])
-AC_SEARCH_LIBS(jent_version,jitterentropy,
-		[AM_CONDITIONAL([JITTER], [true])
-		AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])
-		AC_MSG_NOTICE([Enabling JITTER entropy source])],
-		AC_MSG_NOTICE([Disabling JITTER entropy source]))
+
+AC_ARG_ENABLE(jitterentropy,
+	AS_HELP_STRING([--disable-jitterntropy | --enable-jitterentropy=<path>],
+	[Disable jitterentropy source, or specify its location]),
+	[if [ test -d $enable_jitterentropy ]; then
+		export LDFLAGS+=-L$enable_jitterentropy;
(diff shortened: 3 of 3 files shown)


nhorman:
@abelbeck  could you please test with this, as I think it will have an impact on your use case (it currently doesn't support static linking).  If you try it and are ok wtih that, well and good, but if you would prefer to statically link, then we can add in an option to do that if need be.  I'm thinking that this should probably be merged after then next release, but if you're comfortable with it as is, then I can move more quickly

abelbeck:
@nhorman Please defer this PR after an official rng-tools release.

Our build system currently requires a static jitterentropy build.  The jitterentropy-library does not have any autotools infrastructure, so currently it leverages off your rng-tools package.

I would prefer an option to statically include jitterentropy-library down the road.

nhorman:
@abelbeck I can wait until after I release (which I'll do this afternoon), but I just did some reading and discovered that the configure script as written here will support static libraries.  all you have to do is build jitterentropy, and install the headers/archive files in the library search path, and rng-tools will link against it properly.  I just gave it a shot and it worked fine.  Can you confirm?

abelbeck:
@nhorman Thanks much for the official rng-tools 6.5 release !

Additionally, I want to mention it was a pleasure working with you during the last couple weeks with rng-tools.  rng-tools is an important project.

Back to this PR... When I said a static jitterentropy build, I meant the method rng-tools 6.5 uses where the jitterentropy-library source can be plopped into the rng-tools and build and link statically.  This PR eliminates that method.

Using dynamic libraries is what our project usually does (via Buildroot), but that requires the library source to properly cross-compile, and without autotools infrastructure in jitterentropy-library that may take some hacking.

We are nearing a new release of AstLinux in the next couple weeks, so I won't be able to try to get jitterentropy-library to build as a standalone library for some time.

Ideally rng-tools configure.ac would include a `--with-included-jitterentropy` option to indicate the existence of included jitterentropy-library source and work as rng-tools 6.5 does, and without specifying `--with-included-jitterentropy` jitterentropy is dynamically linked by default.  I've seen this done with other projects.

nhorman:
@abelbeck thank you, its been nice working with you as well.

Regarding this PR, removing jitterentropy source from rng-tools is more or less the purpose of this PR, the desire  to not actually carry the jitterentropy source is what I'm trying to get to (as I don't want to have to keep up with it), like gcrypt, I'd like to just use it if the library is available, and not use it if its not.

I understand the headache of having to modify your packaging, and I don't want to have you do it before a release, so I'm willing to wait on this, but it would be helpful to know what time frame you are working in.

As for future packaging, I feel your pain, I get that it takes a long time to package libraries separately, but in response, I would suggest you, sometime after your next release, just move the building of jitterentropy to the ebuild script for rng-tools in your repository.  That is to say, just add jitterentropy to the SRC_URI list of your ebuild (IIRC ast linux is a gentoo derivative, correct?), and then unpack and build it in your src_unpack and src_build functions (its just one command).  Then configure rng-tools like so:
configure LDFLAGS=-L/path/to/where/you/built/jitterentropy.a

That should allow rng-tools to do a static build without you having to package jitterentropy separately in the near term.  Does that seem like a reasonable solution?

abelbeck:
> and I don't want to have you do it before a release, so I'm willing to wait on this

Thanks, but we will be including rng-tools 6.5 in our next release, so we are good.  You don't need to wait.

> your ebuild (IIRC ast linux is a gentoo derivative, correct?)

No, we cross-compile using Buildroot https://buildroot.org/
(Though the very first AstLinux in 2004 was based on Gentoo)

After a couple of weeks I can look at what it takes to add a jitterentropy package to our Buildroot build system.  It may turn out to be easier for our project to just patch future rng-tools to continue to support internal building of jitterentropy, not sure.

The jitterentropy-library seems quite stable and mature, I would expect it to be minimal maintenance for rng-tools to continue `--with-included-jitterentropy` support ... since you are not including the actual jitterentropy-library source.

nhorman:
ok, copy that I'll merge in the next few days/week then

Regarding packaging jitterentropy with rng-tools, I think it will be pretty
straightforward (thought I'm not familiar with the details of how buildroot
packages its build info).  I'm personally packaging jitterentropy as a
shared library for Fedora (not sure if that helps you out or not long term).
Regarding jadding a --with-included-jitterentropy support, I just got
another request for that, so it appears I'm somewhat in the minority
regarding justpulling it in if the library exists, and I'll pull in that
support with this PR

abelbeck:
> Regarding adding a --with-included-jitterentropy support, I just got
another request for that, so it appears I'm somewhat in the minority
regarding justpulling it in if the library exists

The other PR is asking for something different, `--with-jitter` controlling shared jitterentropy lib support, allowing jitter to be disabled (`--without-jitter`).

Personally, a nit, I think the naming should be `--with-jitterentropy` for the other related PR, matching the name of the shared library package.

nhorman:
but if you allow the enablement of with-jitter, you can imply the inclusion of a local copy of the jitterentropy library by configuring rng-tools with ./configure LDFLAGS=-L./
so that it will pick up whatever jitter library you put in the local source tree

And I agree on the jitterentropy language, thats the name of the upstream project, not jitter

abelbeck:
> but if you allow the enablement of with-jitter, you can imply the inclusion of a local copy of the jitterentropy library by configuring rng-tools with ./configure LDFLAGS=-L./

But why require users to use a hack'y workaround, why not add a `--with-included-jitterentropy` configure option that would provide v6.5-style builds ?

Just think of all the questions you will not need to answer ? :-)

nhorman:
Because 2 options seems awfully confusing to me.  Two options means 4
combinations:

-with-jitterentopy / -with-included-jitterentropy:
     Use the library in the source tree and enable the feature

--with-jitterentropy / null:
    Enable the feature, and only look for jitterentropy in the search path

--without-jitterentropy / -with-included-jitterentropy:
     Don't enable the feature, even though you specified a path to the
included port?  Or override the last --without, and enable the local build?


--without-jitterentropy / null :
    Don't enable jitterentropy at all

The  2nd and 4th entries make.sense to me, and the first does to a degree,
but the 3rd just seems broken.

Fwiw the AC_SEARCH_LIB macro documents the use of LDFLAGS as the method to
search a specific path for a library (useful when a single host targets
mutiple.cross compilation platforms)

That said, perhaps there is a happy compromise here.  What if i just
unilateraly add -L ./jitterentropy to the ldflags in the configure script
prior to running the SEARCH_LIB macro.  You would still have to build it
yourself, but it would prioritize the use of that library if it existed
there.  Basically the --with-included-jitterentopy option would be implied
by the placing/building  of the .a file in that directory.

Thoughts?
Neil

abelbeck:
> That said, perhaps there is a happy compromise here.

There still needs a way to choose static vs. dynamic linking of jitterentropy.

I have defined configure options on 10's if not 100 packages, and configure options are always welcomed, and not all permutations need to be meaningful and simply ignored.

My wish for rng-tools v6.6 configure:

1) `--with-jitterentropy` (default) enable JITTER and use a shared library

2) `--without-jitterentropy` disable JITTER, don't build with jitter support

3) `--with-included-jitterentropy` enable JITTER and use the included (internal) static library (v6.5 style)

4) `--without-included-jitterentropy` (default) a NoOp, ignored.

Action 2. would make 3. ignored

Action 3. would make 1. ignored

Some package's configure will also allow path names in addition to `yes` and `no` for the above to locate the lib's, but such complexity is not required here.

Alternatively some package's configure use `enable-static` and `enable-shared` but I think the above are more specific.

I'll let you decide.

nhorman:
Ok, try this on, the AC_ARG_ENABLE macro allows you to specify a optional argument, so what if we do like I've done in the new commit on this PR, which default enables jitterentropy, allows you to disable it with --disable-jitterentropy, and allows you to specify a location for the jitterentropy library by specifying --enable-jitterentropy=/path/to/library.  --enable-jitterentopy works also, but is a no-op as its already default enabled to just search the standard include path.

nhorman:
I think we're just not going to agree on that.  The entire point of this is
for me to stop carrying the sources for jitterentropy.  If you want to
build a specific version and link it statically to rngd, this allows you to
do that, but its the packagers responsibility to get the jitterentropy
tarball, extract it to a location appropriate  for building, and build it
as you see fit.

abelbeck:
> If you want to build a specific version and link it statically to rngd, this allows you to do that

I don't see how, unless an external patch that adds in Makefile.am bits this PR removes, like here:
https://github.com/nhorman/rng-tools/pull/37/files#diff-c949f93d03f44a4217d7a138f9e2e54aL4

Also it would seem `--disable-jitterentopy` would be defined and patch `HAVE_JITTER` back in somehow. I'm not sure.

Unless I'mm missing something.

nhorman:
Because you won't be building it by building rng-toos.  It will be  your
responsibility as a packager to build jitterentropy on your own, however
you see fit (be that building it as a separate package as I am in Fedora,
or by modifying your rng tools metadata  in your buildroot infrastructure
so that a private copy of libjitterentropy.a is built solely for the
purpose of linking to rngd when you build rng-tools. I should never have
included the jitterentropy source in rng-tools, but I did, and now I'm
trying to clean up that mess.  Starting in 6.5 jitterentropy will be built
exactly like any other library for your use (think gcrypt or Glib or
ncurses, or any other library on your system that your program may use,
thats how jitterentropy is going to be referenced).

As for defining HAVE_JITTER when its disabled, no.  The AC_ARG_ENABLE macro
defines $enable_jitterentropy to one of 4 values:
a) "no" if --disable-jitterentropy is set
b) "yes" if --enable-jitterentropy is set
c) <path> if --enable-jitterentropy=<path> is set
d) nothing if no value is placed on the command line

the AS_IF clause tests for "x$enable_jitterentropy" != "xno" and does the
library search if its true.  If --dsable-jitterentropy is asserted on the
configure command line, then $enable_jitterentropy=no, and so the test
fails, maning the AC_SEARCH_LIBS macro gets skipped entirely.  only if
$enable_jitterentropy is set to something other than "no", do we search for
the library, and only if we find a compatible library do we define
HAVE_JITTER.

abelbeck:
OK, lets agree to disagree and move on...

Please do what you think is best for the rng-tools project.

PR #38, opened 2018-09-26
configure.ac: control jitterentropy support using --with-jitter

gktrk:
Replace automagic detection of jitterentropy DSO with an AC_ARG_WITH
macro so that rngd can be compiled without jitter support even when
the jitterentropy library is available on the system.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,6 +40,13 @@ AC_ARG_WITH([nistbeacon],
 	[with_nistbeacon=check]
 )
 
+AC_ARG_WITH([jitter],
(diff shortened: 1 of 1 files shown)


nhorman:
I've got a pull request that affects this same area thats waiting for a distro release to incorporate.  If you could make this pull request against https://github.com/nhorman/rng-tools/tree/remove-static I'll pull it to there and merge them at the same time.

nhorman:
I'm taking care of this in the remove-static pull request [here](https://github.com/nhorman/rng-tools/pull/37)

PR #39, opened 2018-09-30
 Obey restrictions on x86_rdrand_bytes usage 

lonwillett:
Make xread_drng obey the restrictions imposed by x86_rdrand_bytes.
Otherwise, problems can happen (e.g. the call to discard_initial_data results in a smashed stack, which is how I encountered this problem).

### DIFF
--- a/rngd_rdrand.c
+++ b/rngd_rdrand.c
@@ -246,7 +246,14 @@ int xread_drng(void *buf, size_t size, struct rng *ent_src)
 	if (ent_src->rng_options[DRNG_OPT_AES].int_val)
 		return xread_drng_with_aes(buf, size, ent_src);
 
-	x86_rdrand_bytes(buf, size);
+	/* NB: x86_rdrand_bytes might overrun end of buffer, if not a multiple of 8 */
+	if (size > 7)
+		x86_rdrand_bytes(buf, (size&~7));
+	if ((size&7) != 0) {
+		unsigned char tempbuf[8];
+		x86_rdrand_bytes(tempbuf, (size&7));
+		memcpy((unsigned char *)buf+(size&~7), tempbuf, (size&7));
+	}
 	return 0;
 }
 


abelbeck:
@lonwillett You mention it fails on a "call to discard_initial_bytes" ... do you mean `discard_initial_data()` in rngd.c ?
https://github.com/nhorman/rng-tools/blob/master/rngd.c#L627

Do you see other places where a `x86_rdrand_bytes()` overrun could cause problems, or just use a larger `tempbuf` in `discard_initial_data()` in rngd.c ?

abelbeck:
Hmmm, `FIPS_RNG_BUFFER_SIZE` is 2500 which is not a multiple of 8, (currently, pre-PR) does that mean 2500-2503 could be written to by xread_drng ?

abelbeck:
I looked, and this occurrence of `x86_rdrand_bytes()` that @lonwillett fixes in this PR seems to be only occurrence that can have a non-mod-8 value of `size`.

Question, in the line
```
memcpy(tempbuf, buf+(size&~7), (size&7));
```
shouldn't the `void *buf` be cast as `unsigned char*` for the proper pointer arithmetic ?

abelbeck:
@lonwillett You have the `memcpy` pointer arguments reversed.

lonwillett:
@abelbeck 

> do you mean discard_initial_data

Yes.

> shouldn't the void *buf be cast as unsigned char* for the proper pointer arithmetic ?

Yes, I was lazy, and AFAIK void* arithmetic is still not portable. Change made.

> You have the memcpy pointer arguments reversed.

Oops. That's embarrassing. Now fixed.

nhorman:
@abelbeck runs rngd on a 32 bit machine, its a quite popular embedded platform, this fix should definitely apply there as well.

lonwillett:
This also fixes 32-bit platforms; it just slightly overdoes it. It makes sure the buffer size is a multiple of 8 (as required for 64-bit platforms). A 32-bit platform only requires a multiple of 4.

abelbeck:
Actually, our project uses both 32-bit and 64-bit toolchains ... with `FIPS_RNG_BUFFER_SIZE` set to 2500, 32-bit compilers would be immune to this issue, for some reason it does not crash (for me) on 64-bit systems.  Our compiler must leave 4 bytes of padding after the non-mod-8 stack based arrays for word alignment.

Reguardless this fix should apply to everyone for completeness.  Thanks @lonwillett !

PR #44, opened 2019-02-13
Fixed else branch for random_add_entropy

ya-isakov:


nhorman:
I see now what you mean regarding the else clause, which is fine, but I'm not loving the reactions to a failed entropy count read.  In fact, I'm not sure how I let the two RNDGETENTCNT ioctls in there.  We should only have one in the else clause of the RNDADDENTCNT ioctl, as we should never have a need to get the entropy count in the if clause (because the ADDENTCNT ioctl failed, either due to an internal failure, or because the output file isn't the random device in the kernel and won't respond to the GETENTCNT ioctl in anyway

This:
if (ioctl(random_fd, RNDGETENTCNT, &ent->ent_count) != 0)
                                message(LOG_DAEMON|LOG_ERR, "Failed to get Entropy count!\n");
                        ent->ent_count *= 8; /*adjust for the kernel shift*/
Should just be removed entirely (and the *=8 is erroneous).

I'm also not loving how we react to a failure.  It should never happen in the else clause (the only error we can get is EFAULT if the put_user fails), but if it does fail, we should probably return 0, to ensure that we try again immediately

ya-isakov:
I've closed this, as per comment in another discussion https://github.com/nhorman/rng-tools/commit/a88c630acb6e1c0403a74990b8f4e63468819fe7#r32309185

PR #49, opened 2019-04-05
configure.ac: fix autoreconf

ffontaine:
Use foreign instead of gnu in AM_INIT_AUTOMAKE otherwise autoreconf will
fail on:
Makefile.am: error: required file './README' not found

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@ AC_INIT(rng-tools, 6.7, [Neil Horman <nhorman@tuxdriver.com>])
 AC_PREREQ(2.52)
 AC_CONFIG_SRCDIR([rngd.c])
 AC_CANONICAL_TARGET
-AM_INIT_AUTOMAKE([gnu])
+AM_INIT_AUTOMAKE([foreign])
 AC_CONFIG_HEADERS([rng-tools-config.h])
 AC_CONFIG_MACRO_DIRS([m4])
 


nhorman:
why would it fail?  There is a README in the tree.

ffontaine:
There is no README in rng-tools-6.7.tar.gz or in the tree, README file has been renamed in README.md since version 6.1 and https://github.com/nhorman/rng-tools/commit/5eaecc3fd6fead8fe83233fc070a85301bdc7b89. Thus, since this commit, AM_INIT_AUTOMAKE fails because by specifying gnu, we asks automake to check for compliance to the GNU standards (ChangeLog, AUTHORS, README, ...)

nhorman:
I'm sorry, I wasn't clear.  You are correct, there is no README file checked into the git tree directly, but it is generated at configure time via the autogen script, when we run automake with the --add-missing option.  If you are manually running the automake command, you need to use that option (as per the autogen.sh script)

ffontaine:
I would like to avoid using --add-missing option. I'm updating the current version of rng-tools in buildroot and we have a standard autoreconf mechanism for all packages.

And other option would be to include a configure script in your release tarballs so I wouldn't have to call autoreconf (as it was done in the tarballs of the sourceforge project).

The latest option for me is to add this PR's commit in buildroot but we tried as much as possible to avoid non-upstreamable patches.

abelbeck:
@ffontaine For our Buildroot rng-tools we use:
```
define RNG_TOOLS_CONFIGURE_PREFLIGHT
	( cd $(@D) ; \
	  $(HOST_CONFIGURE_OPTS) \
	  ./autogen.sh ; \
	)
endef
RNG_TOOLS_PRE_CONFIGURE_HOOKS += RNG_TOOLS_CONFIGURE_PREFLIGHT
```

ffontaine:
Thanks for your input @abelbeck, this is indeed a fourth option.

If you think that replacing gnu by foreign in AM_INIT_AUTOMAKE call is definitely a bad idea, I can update buildroot with this solution (or you could even contribute this proposal yourself, we're always happy to have new contributors).

abelbeck:
@ffontaine Feel free to use the `RNG_TOOLS_CONFIGURE_PREFLIGHT` solution above ... we use an older version of Buildroot, but I think this would still work upstream.

nhorman:
I'm not opposed to making the change necessarily, just trying to understand why its needed, as modifying this option reduces the level of compliance to gnu build standards.  I've done some quick comparisons though and the currently tree doesn't change the build at all, so I'd be comfortable taking this.  Would you prefer that, or to use @abelbeck solution?

ffontaine:
I would prefer that you accept the PR. @abelbeck solution is ok but it is less elegant than just doing `RNG_TOOLS_AUTORECONF=YES` in the buildroot's rng-tools.mk file.

PR #56, opened 2019-04-17
Fix "make distcheck"

rubenk:


nhorman:
I'm going to need a better commit log than that.  Please add more detail as to what exactly you are fixing

rubenk:
I've amended the commit message.

nhorman:
thank you, that works

PR #66, opened 2019-06-06
Import yocto fixes for 6.6

nicola-lunghi:
This pull request is to import yocto fixes for rng-tools 6.6

see here
https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-support/rng-tools

some of those changes came from gentoo

https://packages.gentoo.org/packages/sys-apps/rng-tools
https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-apps/rng-tools

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -47,6 +47,13 @@ AC_ARG_WITH([pkcs11],
 	[with_pkcs11=check]
 )
 
+AC_ARG_WITH([libargp],
+	AS_HELP_STRING([--without-libargp],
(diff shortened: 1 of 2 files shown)


nhorman:
Ok, I've tested these, and they look ok. I would appreciate it though if you could fix up the commit messages.  Theres no need for the Patch released under the GPLv2 language in the rdrand patch, and none of the patches need the Upstream-status: pending note, as that will be out of date as soon as I merge it.

nicola-lunghi:
Ok thanks I will fix it

PR #72, opened 2019-11-06
Remove SIGTERM/SIGINT handling

cgwalters:
The previous code had bugs, see:
62fbff0ac807e07f4f21f58c0f15bfe5e9c8b55e
The version in RHEL doesn't have this patch, and we're hitting
shutdown bugs in RHEL CoreOS.

There's no actual good reason though to have a `SIGTERM` handler
at all; doing things like `pthread_join()` is pointless becuase
the kernel will take care of removing all resources.

In fact it's really an anti-pattern to do this, because if every
process has these types of things, it makes shutdowns much less
efficient as the process needs to be fully scheduled in rather
than just having the kernel take care of it.

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -65,7 +65,6 @@ int kent_pool_size;
 bool am_daemon;				/* True if we went daemon */
 bool msg_squash = false;		/* True if we want no messages on the console */
 bool quiet = false;			/* True if we want no console output at all */
-volatile bool server_running = true;	/* set to false, to stop daemon */
 
 bool ignorefail = false; /* true if we ignore MAX_RNG_FAILURES */
 
@@ -542,8 +541,6 @@ static int update_kernel_random(struct rng *rng, int random_step,
 
 	for (p = buf; p + random_step <= &buf[FIPS_RNG_BUFFER_SIZE];
 		 p += random_step) {
-		if (!server_running)
-			return 0;
 		rc = random_add_entropy(p, random_step);
 		if (rc == -1)
 			return 1;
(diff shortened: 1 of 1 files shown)


cgwalters:
xref https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/688

nhorman:
I don't agree in principle  with the notion of not having an application gracefully release its resources on shutdown.

Instead of removing that code, why not just open a bug with RHEL to backport the needed change to fix the problem?  You're going to have to do that anyway to get any change made here today into RHEL

cgwalters:
> I don't agree in principle with the notion of not having an application gracefully release its resources on shutdown.

Can you elaborate more?  

I think this isn't just bikeshedding, it's gets to a really fundamental principle of systems design.  See also e.g. https://en.wikipedia.org/wiki/Crash-only_software - that strongly influenced ostree.  I spent a lot of time designing it so that it's totally safe to be interrupted at any point.  It was only recently that I was forced to add a `SIGTERM` handler for a corner case, see https://github.com/ostreedev/ostree/pull/1049

cgwalters:
> You're going to have to do that anyway to get any change made here today into RHEL

Well...we could also change the systemd unit easily in RHCOS to use `TimeoutStopSec=1` too.

nhorman:
Sure, we could do that too, but thats not my point.  My point is that this is already fixed here in the upstream project, and regardless of what you want the fix to be, it will need to be backported into RHEL.  So why not take the path of least resistance and open a RHEL bz requesting that the existing fix be backported?

cgwalters:
Sure, done https://bugzilla.redhat.com/show_bug.cgi?id=1769916

nhorman:
thank you, I'll take care of it ASAP

PR #84, opened 2020-02-28
rngd.c: only print the success message when initialize

Mingli-Yu:
Only print the message for the entropy sources which
initialized successfully to avoid misleading users.

That makes sense as after check the available entropy
sources, there is logic to check whether there is at
least one available entropy source exist as below:
if (!ent_sources) {
     message(LOG_DAEMON|LOG_ERR,
             "can't open any entropy source");
     message(LOG_DAEMON|LOG_ERR,
             "Maybe RNG device modules are not loaded\n");
     return 1;
}

That's to say, only printing the message for the entropy
sources which initialized successfully is safe.

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -795,9 +795,8 @@ int main(int argc, char **argv)
 				ent_sources++;
 				ent_src->fipsctx = malloc(sizeof(fips_ctx_t));
 				fips_init(ent_src->fipsctx, discard_initial_data(ent_src));
-                                message_entsrc(ent_src, LOG_DAEMON|LOG_INFO, "Initialized\n");
+                                message_entsrc(ent_src, LOG_DAEMON|LOG_INFO, "Initialized successfully\n");
 			} else {
-                                message_entsrc(ent_src, LOG_DAEMON|LOG_ERR, "Initialization Failed\n");
 				ent_src->disabled = true;
 			}
 		}


nhorman:
2 things:

1) You're patch changes don't match what you say in your changelog.  You say you want to print print messages only for for entropy sources that initialize successfully, but your patch still prints  a message when the initialization of an entropy source fails

2) I disagree that printing failure messages is misleading, it is reporting exactly the truth of the result of a init call.  The confusion here surrounds the understanding that rngd supports multiple entropy sources, and only one needs to initialize properly for rngd to function correctly.  To fix that, I would suggest that we add documentation to the rngd man page under the ENTROPY SOURCES section documenting both the fact that we have multiple sources, which may or may not be available, and that only one needs to be present and functional for the daemon to work

Mingli-Yu:
Thanks nhorman for your quick response!
1, we only print message when ent_src->init(ent_src) return 0 with my patch, it should just print message when initialize succesfully.
2, Though add the explanation to the doc makes sense, but frankly it is clear when just check the output of "systemctl status rngd". Moreover, we maybe only care the entropy sources which initialized succefully. It's okay to silence the message for those entropy sources which failed to initialize.

nhorman:
I see what you're saying here, in that you're supressing/removing the message when the initialization of an entropy source fails, but I don't agree with that, as it denies the user useful information when they expect a given entropy source to function properly.  I would really rather just document the idea of expected failures in the man page

PR #90, opened 2020-03-30
Nistbeacon fixes

gktrk:
Refresh the record every 60 seconds.

Note that this mechanism breaks the rngd test mode (`--test`). The breakage is originally introduced by 6f570a653a0dae833c014d336a60a0c867613982, not by this PR. This is because the test mode also uses a handler for SIGALRM, overriding nistbeacon.

Another note is the possible interference by libcurl. In certain configurations it may use the same SIGALRM mechanism to implement timeouts for connections (see https://github.com/curl/curl/blob/4a4b63daaa01ef59b131d91e8e6e6dfe275c0f08/lib/hostip.c#L723).

### DIFF
--- a/rngd_nistbeacon.c
+++ b/rngd_nistbeacon.c
@@ -64,7 +64,7 @@
 #define NIST_BUF_SIZE 64 
 #define NIST_CERT "/home/nhorman/Downloads/beacon.cer"
 
-bool get_new_record = false;
+bool get_new_record = true;
 
 static int get_nist_record(struct rng *ent_src);
 
--- a/rngd_nistbeacon.c
+++ b/rngd_nistbeacon.c
@@ -689,7 +689,9 @@ static int get_nist_record(struct rng *ent_src)
 		goto out;
 	}
 
-        
+	/* Wait 60 seconds before fetching the next record */
+	alarm(60);
+
 	rc = 0;
 
 out:


nhorman:
you're right, I totally forgot about the test mode alarm handler, and resetting the alarm.  I think we can't make test mode and nistbeacon with a signal-timed record fetch play nice together though, not without creating some sort of centralized alarm handler.

I think the better approach here is to disable nist_beacons alarm handler alltogether, and fetch records on a 60 second timed basis using clock_gettime

nhorman:
https://github.com/nhorman/rng-tools/commit/a8b0d0512e0e0f201479df2aaa90edb891567876

That change works for me, can you test it out and confirm?

gktrk:
> [a8b0d05](https://github.com/nhorman/rng-tools/commit/a8b0d0512e0e0f201479df2aaa90edb891567876)
> 
> That change works for me, can you test it out and confirm?

:+1:

The only downside I'm seeing is the the added system call overhead but I suppose it's not really a big deal.

nhorman:
I'll update it to use CLOCK_MONOTONIC_COARSE, so that, on systems which support it, your syscall time per check will get reduced.  On my system here, that reduces the syscall time from 17ns to 6ns.

nhorman:
fixed with https://github.com/nhorman/rng-tools/pull/91

PR #92, opened 2020-05-11
Add systemd service file

tpgxyz:
This PR add systemd service file. It is not installed by default.
This PR also helps to reslove this https://github.com/dracutdevs/dracut/pull/791

### DIFF
--- a/rngd.service
+++ b/rngd.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Hardware RNG Entropy Gatherer Daemon
+ConditionVirtualization=!container
+
+[Service]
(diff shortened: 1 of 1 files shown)


nhorman:
I can certainly add this, but usually distributions assemble these files on their own, I'm not sure why we need it here.  Also, your change doesn't actually install anything, it just creates the file in the repository.  Was it your intent to augment the makefile to install the service on the running system?

tpgxyz:
@nhorman I may add install to Makefile, but distributions may not use it at all, so it treat it as a good example for packagers

nhorman:
ok, I'm fine with that

Issue #117, opened 2021-01-13
jitter source makes rngd hangs at 100% cpu usage


xse:
Hello,
A recent update of rng-tools introduced a 100% CPU usage from rngd, this is related to that jitter source and `rngd -f -x jitter` fixes the issue.
I'm using a raspberry pi (armv6) machine running archlinuxarm, the same issue has been reported on raspberry pi zero, 1b, 3b and 4.

An strace leads to believe that rngd gets stuck in an endless loop of `clock_gettime64` which does not happen when `-x jitter` is used.

Here's a sample strace output:
```
write(2, "Initializing available sources\n", 31Initializing available sources
) = 31
openat(AT_FDCWD, "/dev/hwrng", O_RDONLY|O_NOCTTY) = 3
read(3, "\23IJ4\255\330\340\312z\tn\367&\270g\243", 16) = 16
read(3, "Nc)`", 4)                      = 4
read(3, "\347\3\270\334", 4)            = 4
write(2, "[hwrng ]: Initialized\n", 22[hwrng ]: Initialized
) = 22
rt_sigaction(SIGUSR1, {sa_handler=0x46a7e0, sa_mask=[USR1], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x76b02120}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
openat(AT_FDCWD, "/proc/sys/crypto/fips_enabled", O_RDONLY) = -1 ENOENT (No such file or directory)
clock_gettime64(CLOCK_REALTIME, {tv_sec=1610569934, tv_nsec=579021205}) = 0
clock_gettime64(CLOCK_REALTIME, {tv_sec=1610569934, tv_nsec=581817181}) = 0
clock_gettime64(CLOCK_REALTIME, {tv_sec=1610569934, tv_nsec=583609165}) = 0
... (gets stuck doing endless clock_gettime64)
```

The same thing arise when using `rngd --list` where the program gets stuck mid list and has to be killed.
I'm afraid i can't give you much details, `--debug` doesn't give any useful information and there isn't anything useful in the logs.
I'm not confident in my capacity to pinpoint and tell you exactly where the problem is, but i'mhappy to assist however i can if you need more information, i'm willing to try anything and/or give you an ssh access to somewhere where you could replicate the issue.

Have a good day!

nhorman:
[...]

PR #122, opened 2021-04-23
rngd.c: don't encourage to use the jitter entropy source

Mingli-Yu:
On some systems [1], the jitter entropy source initialization makes rngd
hang at 100% cpu usage.

So only fall back to use this jitter entropy source when there is no
other available entropy source.

[1] https://github.com/nhorman/rng-tools/issues/117

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -822,7 +822,7 @@ int main(int argc, char **argv)
 	int ent_sources = 0;
 	pid_t pid_fd = -1;
 	double test_time;
-	struct rng *ent_src;
+	struct rng *ent_src, *ent_src_jitter;
 
 	openlog("rngd", 0, LOG_DAEMON);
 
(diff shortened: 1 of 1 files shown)


nhorman:
This isn't acceptable.  Many users enable the jitter entropy source in parallel with other sources so that they have a fall back in the event that other sources fail.  This change decides at startup weather jitter should be enabled and doesn't allow for that fallback condition.  Additionally, we already have a mechanism to de-prioritize jitter at run time if faster sources are available (it works for the NIST beacon as well).  The problem you are having is that you are hitting all the high cpu usage during rngd initialization, which has to be done before using the jitter source at all.

On a given platform, you should have some level of insight as to what entropy sources are available before boot, and weather or not those systems will be impacted with high cpu utilization if jitter is enabled.  For those systems I would suggest writing a patch to augment  the systemd startup script, something like this:
EnvironmentFile=-/some/env/file
ExecStartPre=-/path/to/detectionscript
ExecStart=/usr/sbin/rngd -f $JITTER_EXCLUDE

Where detectionscript is some script you as a system builder can optionally write that detects at runtime the system you are working on, along with any other entropy sources that may be useable, and correctly sets $JITTER_EXCLUDE to an empty string or "-x jitter" accordingly, and writes that into /some/env/file

That way at system boot you can determine if you want to completely exclude jitter or not based on the availabilty of other entropy sources on systems that will encounter high cpu usage if jitter is enabled

Mingli-Yu:
@nhorman ,
Thanks very much for your clarification!
If so, I think it's necessary for us to address the jitter entropy source hang issue, especially on some systems which use the internal timer during the jitter entropy source initialization. Please help to check if the https://github.com/nhorman/rng-tools/pull/123 is acceptable.

PR #123, opened 2021-04-25
rngd_jitter.c: don't set the CPU affinity mask

Mingli-Yu:
For the jitter entropy source, each task thread will create an internal
counter timer thread when the system clock resolution is under 5MHz.

But it will introduce high cpu usage for a long time and also make random
data generate too slow if sets the CPU affinity mask of the internal counter
timer thread.

So remove the CPU affinity mask setting to avoid lots of context switch
and too high cpu load for a long time.

Signed-off-by: Zhantao Tang <zhantao.tang@windriver.com>

### DIFF
--- a/rngd_jitter.c
+++ b/rngd_jitter.c
@@ -252,11 +252,6 @@ static void *thread_entropy_task(void *data)
 		me->slptm.tv_nsec = 0;
 	}
 
-	/* fill initial entropy */
-	CPU_ZERO(&cpuset);
-	CPU_SET(me->core_id, &cpuset);
-	pthread_setaffinity_np(pthread_self(), CPU_ALLOC_SIZE(me->core_id+1), &cpuset);
-
 	tmpbuf = malloc(me->buf_sz);
 	if (!tmpbuf) {
 		message_entsrc(me->ent_src,LOG_DAEMON|LOG_DEBUG, "Unable to allocate temp buffer on cpu %d\n", me->core_id);


nhorman:
No, this seems very wrong.  At best you're fixing your own problem at the expense of other cpus which are more powerful.  At worst, you are either creating more contention for multiple threads, or simply confusing the code (in the event the scheduler accurately spreads the threads over your cores).

Do you have data showing this change solves the problem you are seeing?  If so, I would like to see it please.  If it does in fact solve the problem, without causing undue problems for more powerful systems, we can consider this, but I would like to look at conditions in which we actively detect which systems need to have affinity disabled

Mingli-Yu:
@nhorman,
I have verified the above solution and it proves to solve the issue.  BTW, after the affinity disabled as above, the jitter entropy source initialization finish within 1 second("rngd -f  -x hwrng -x rdrand" return immediately). And if not disable, "rngd -f  -x hwrng -x rdrand" hang there and the cpu load usage take 100% for about 30 minutes. 

And you can reproduce the issue easily in any bsps after you force to use the internal timer with below patch as below for jitterentropy-library(git://github.com/smuellerDD/jitterentropy-library.git) Note: on some boards, no need force to use the internal timer as they have high resolution timer, but to simulate the behaviour on some low  resolution timer, we can unify to use the below patch to let the board use the internal timer.
diff --git a/jitterentropy-base.c b/jitterentropy-base.c
index 6dbb484..3c11672 100644
--- a/jitterentropy-base.c
+++ b/jitterentropy-base.c
@@ -676,7 +676,7 @@ static int sha3_tester(void)
  * that no suitable time source is available.
  ***************************************************************************/

-static int jent_force_internal_timer = 0;
+static int jent_force_internal_timer = 1;

 /**
  * Timer-replacement loop
@@ -1387,7 +1387,8 @@ int jent_entropy_init(void)
        if (sha3_tester())
                return EHASH;

-       ret = jent_time_entropy_init(0);
+       //ret = jent_time_entropy_init(0);
+       ret = 1;

 #ifdef JENT_CONF_ENABLE_INTERNAL_TIMER
        jent_force_internal_timer = 0;

nhorman:
Ok, I see now.  This is still not a good solution to the problem you are facing, but it makes more sense now.  I presume you are building your system with the latest jitterentropy library?  Apparently last november, smuller added the ability to use a pthread as a high resolution timer, which I was unaware of (though I should have been).  I expect this was done in response to a discovery made that on the Azure cloud some vms were provisioned in such a way that the rdtsc instruction was virtualized to provide a very coarse time granularity which was not suitable to jitterentropy usage.

So the first question I would have for you is, does your hardware have an rdtsc instruction, or does it support clock_gettime(CLOCK_REALTIME)?  If it does, please just undefine JENT_CONF_ENABLE_INTERNAL_TIMER in jitterentropy.h and rebuild.  You have no need to use the internal pthread timer emulation at all, and all of this just goes away.  Its still a high load entropy source, but any additional load you see from thread contention will not be present.

If you do not have a hardware enabled HR timer (in which case I would be interested to know what hardware you are using), thenperhaps we need to do something, though looking at it, it seems like the solution you are proposing isn't correct.  I would say that the problem is actually in jitterentopy, not rngd.  rngd explicitly wants to set affinity per thread to each cpu to ensure that each cpu is generating entropy.  if you use the internal timer thread, then its just inheriting the parent threads affinity mask, where what you want is (I expect) to have the timer thread expressly set its affinity to all cpus.  Thats probably the patch we're looking for (though I would strongly suggest just disabling the INTERNAL_TIMER define first, as I expect you have hardware support for high resolution timers.

Mingli-Yu:
@nhorman,
I use the latest jitterentropy-library 3.0.2 (git://github.com/smuellerDD/jitterentropy-library.git).

The hardware(nxp-s32g2xx) I use has the rdtsc instruction and also support clock_gettime(CLOCK_REALTIME). And if undefine JENT_CONF_ENABLE_INTERNAL_TIMER, then jitter entropy failed to initailize as jent_entropy_init returns not 0 as  jent_time_entropy_init(0) return not 0 and no internal phtead timer available after we undefine JENT_CONF_ENABLE_INTERNAL_TIMER as below.
`JENT_PRIVATE_STATIC
int jent_entropy_init(void)
{
        int ret; 

        if (sha3_tester())
                return EHASH;

        ret = jent_time_entropy_init(0);
      #ifdef JENT_CONF_ENABLE_INTERNAL_TIMER
        jent_force_internal_timer = 0; 
        if (ret) {
                ret = jent_time_entropy_init(1);
                if (!ret)
                        jent_force_internal_timer = 1; 
        }
#endif /* JENT_CONF_ENABLE_INTERNAL_TIMER */

        return ret; 
}
`

I don't quite understand why do we need to set affinity per thread to each cpu to ensure that each cpu is generating entropy as you mentioned. After disable the affinity mask, it not only address the hang issue, but also can make the entropy data collect quickly and generate randomly.

nhorman:
As I've said before, I want to ensure that each cpu is generating entropy instead of a single cpu that is potentially just aliasing entropy gathering from scheduler interference.  Regardless, setting thread affinity in an application shouldn't be cause for a supporting library to degrade its performance so significantly.  The way we call jent_entropy_init, the timer thread in the jitter library should be affined to all cpus, so if we simply reduce the jitter thread count to 1 here, the problem should be solved.

In either case, if disabling the internal timer causes errors for you, it would appear that your rdtsc instruction is too coarse to use effectively for jitter measurements, which is unfortunate.  Whats the specific error that it returns, just for my own education.

Regardless of the error however, I'm still not seeing how your patch would do anything, if we simply reduced the number of threads we create at startup (i.e. -O jitter:num_threads:1)

Mingli-Yu:
@nhorman,
As you mentioned, I reduce the number of threads to 1 and it proves my patch still can solve the issue. BTW, it still hang there if reduce the num_threads to 1 without my patch.

nhorman:
you're missing the point, reducing the number of threads to 1 should also prove your patch isn't needed.  Though i'm having trouble parsing your statement above.  You say both that:
" I reduce the number of threads to 1 and it proves my patch still can solve the issue"
and
"BTW, it still hang there when reduce the num_threads to 1 without my patch"

So does reducing the number of threads to 1 help or not?  It sounds like it doesn't, in which case I have questions as to whats going on here.  

your patch should be improving performance by allowing threads to run on multiple different cpus, thereby avoiding scheduling contention that might otherwise occur when each cpu has at least one high-load thread running on it, right?  The current code executes jent_entropy_init from the main process task, meaning it should have an affinity mask of all cpus, which doesn't help if every cpus has a high load thread on board that is contending for time.  However, it should help if there are multiple cpus, of which only 1 has a high load thread running on it, as the timer thread should be able to run on any other cpu.

But it sounds like you're saying it doesn't help, which suggests to me that something else is going on.  Specfically that, despite having other cpus available on which to run, both the timer thread and the created jitter thread are contending for time on the same cpu, which seems wrong.  Would you agree?

So it seems to me like your patch might make the problem go away, but we don't fully understand what the problem is, and i would like to do that before we take the affinity code apart out of hand.


I also note above that you posted your output with 1 thread enabled, and indicated that it hangs after that output.  from your debug logs I see that it took a jitter thread about 3 seconds to produce its initial entropy, which isn't unreasonable for  that entropy source.  However, then it very clearly says :
[jitter]: Initialized
at the end, indicating that jitterentropy has finished initalization, at which point you say it "hangs".  What are you expecting to happen next?  If jitter is the only active entropy source, that printout marks the end of entropy source initialization, at which point rngd, should complete setup, and go to sleep on a read of the /dev/random, until the kernel wakes it up requesting more entropy.  If thats what you're seeing, are you sure that you're getting a hang at all, or perhaps you are seeing a hang in some other location?

Mingli-Yu:
@nhorman,
What I mean is:
Even reduce the number of threads to 1, it still hang here if use the internal timer. And with my patch, the issue gone even when reduce the number of threads to 1.

nhorman:
@Mingli-Yu I understand exactly what you mean.  You're saying that, even if you reduce the number of threads we create for the jitter entropy source, you still experience a hang.  The only way you've removed the hang is to apply your patch that removes all affinity restrictions from the jitter entropy threads.

Now please take a moment to understand what I am saying.  What I am saying is twofold:

1) I am saying that It is my understanding that the way your patch solves the issue, the same result should be achievable by reducing the thread count to 1.  I assert this because it is my understanding that your patch avoids the hang/extreme latency by removing the affinity restrictions on the jitter threads so that threads can run on different cpus, avoid the delay created by two high load threads competing for a single cpu.   I am saying the same result should be achievable by reducing the thread count so that the timer thread can migrate itself to a different cpu.  I am saying that is not happening, and I want to understand why before I accept a patch that I don't like the behavior of, just because it makes your problem go away.  Instead of continually telling me that the problem goes away with your patch, please help me understand why the reduction of thread count isn't also solving the problem (more on this below).

2) I am also saying that, based on this log data you provided:
[jitter]: Limiting thread count to 1 active cpus
[jitter]: JITTER starts 1 threads
[jitter]: CPU Thread 0 is ready
[jitter]: Initializing AES buffer
[jitter]: JITTER thread on cpu 0 wakes up for refill
[jitter]: xread_jitter requests 16 bytes from pipe
[jitter]: jent_read_entropy time on cpu 0 is 3.249776000000e+00 sec
[jitter]: Writing to pipe
[jitter]: DONE Writing to pipe with return 16535
[jitter]: JITTER thread on cpu 0 wakes up for refill
[jitter]: xread_jitter gets 16 bytes
[jitter]: xread_jitter requests 128 bytes from pipe
[jitter]: xread_jitter gets 128 bytes
[jitter]: xread_jitter requests 16535 bytes from pipe
[jitter]: xread_jitter falls back to AES
[jitter]: Enabling JITTER rng support
[jitter]: xread_jitter requests 4 bytes from pipe
[jitter]: xread_jitter falls back to AES
[jitter]: xread_jitter requests 4 bytes from pipe
[jitter]: xread_jitter falls back to AES
[jitter]: Initialized
[snip]


All the jitter threads should at the point that "[jitter]: Initialized" gets printed, should be blocked on a pthread_cond_wait operation, and so should not be executing.  As such I don't see how the jitter threads could be causing a hang.  I would like to understand where exactly this hang is occuring before I accept a patch that just makes your problem go away.

Now, regarding some information as  to why this might be happening.  Looking at the jitterentropy code, I was under the impression that the internal timer thread was being created when the library was initialized (i.e. it was created from the main thread), and so it was inheriting the main threads affinity mask of all cpus.  However, that is not the case, a unique thread is being created for each task attmpting to grab entropy, which is inefficient for two reasons.  1, it means lots of thread creation and destruction, which we don't need, and 2, it means the timer threads affinity mask is inherited from the calling thread, which by necessity is currently set to a single cpu, leading to the problem.  What jitterentropy should be doing  is enforcing that the timer thread not be constrained to the cpu that the calling thread was constrained to.....maybe.  I need to think about this some more.  On one hand it makes sense to inherit the calling threads affinity mask, but at the same time, that is guaranteed to be a pessimal case here.  The whole notion of an internal timer thread is really just a lousy idea, especially if its higher resolution than the rdtsc instruction that the hardware provides.

Mingli-Yu:
@nhorman,
Many thanks for your clarification! Glad we are at the same page what the problem is.

As I observed the main thread hangs when call "request = read(pipefds[0], &bptr[size-need], need);"
| init_jitter_entropy_source
   | xread_jitter(aes_buf, tdata[0].buf_sz, ent_src);
       | request = read(pipefds[0], &bptr[size-need], need);

And the entropy thread hangs when call "ret = jent_read_entropy(me->ec, tmpbuf, me->buf_sz);"
| thread_entropy_task
   | ret = jent_read_entropy(me->ec, tmpbuf, me->buf_sz);

Look forward your help to guide how to address the issue, thanks!

nhorman:
Given this conversation:
https://github.com/smuellerDD/jitterentropy-library/issues/37
Regardless of the solution we land on, this really needs to be solved internally to the jitterentropy library.  Specifically, if jitterentropy needs a software defined free running counter due to the coarseness of the systems hardware timer, it needs to run in such a way that it minimizes contention with the application (rngd) operation.  Perhaps thats in the form of some affinity configuration methods, or perhaps libjitterentropy just need the ability to punt the operation back to the application (i.e. provide a method to register a thread that does free running counter work).  I'm not sure, but lets continue the converation on the jitterentropy issue above

PR #127, opened 2021-05-09
add option for passing path to pkcs11 engine

jpalus:


nhorman:
This option already exists as -O pkcs11:engine_path.  You can set the default at build time, and override it at run time

jpalus:
The pull request was made from a package maintainer perspective so that distribution packaged rng-tools has default path matching location of engine as packaged in opensc in same distribution without need to override with an option.

jpalus:
I already modify DEFAULT_PKCS11_ENGINE but was hoping for a cleaner solution. Such paths are usually customizable through configure options and clearly indicated in {{configure --help}}.

PR #128, opened 2021-05-18
Add support for ARM v8.5A "rndr" instruction

ozbenh:
This adds support for the RNDR instruction as an entropy source
when available.

There should be no need to run it through AES.

Tested using qemu

Signed-off-by: Benjamin Herrenschmidt <benh@amazon.com>

### DIFF
--- a/rngd_rndr.c
+++ b/rngd_rndr.c
@@ -1,0 +1,193 @@
+/*
+ * Copyright (c) 2021, Benjamin Herrenschmidt and
+ *                     Balbir Singh, Amazon.com, Inc. or its affiliates
+ *
+ * Loosely based on rngd_darn.c: Copyright (c) 2017, Neil Horman
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms and conditions of the GNU General Public License,
+ * version 2, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin St - Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+#define _GNU_SOURCE
+
+#ifndef HAVE_CONFIG_H
+#error Invalid or missing autoconf build environment
+#endif
+
+#include "rng-tools-config.h"
+
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/auxv.h>
+
+
+#include "rngd.h"
+#include "rngd_entsource.h"
+#include "ossl_helpers.h"
+
+/* Kernel headers may be too old */
+#ifndef HWCAP2_RNG
+#define HWCAP2_RNG          (1 << 16)
+#endif
+
+static struct ossl_aes_ctx *ossl_ctx;
+static unsigned char key[AES_BLOCK];
+static unsigned char iv_buf[AES_BLOCK];
+
+#define CHUNK_SIZE AES_BLOCK * 8
+static unsigned char aes_buf[CHUNK_SIZE];
+static size_t aes_buf_pos;
+#define REKEY_BITS 8
+static int rekey_count;
+
+/*
+ * Runs the get_rndr instruction, returns false on failure
+ */
+static bool get_rndr(uint64_t *rndr_val)
+{
+	bool ok;
+	int i;
+
+	/* Let's try 10 times, that should be enough, before we give up */
+	for (i=0; i < 10; i++){
+		asm volatile("mrs %0, s3_3_c2_c4_0\n"
+			     "\tcset %w1, ne\n"
+			     : "=r" (*rndr_val), "=r" (ok)
+			     :
+			     : "cc");
+		if (ok)
+			break;
+		/*
+		 * The spec implies we should wait some milliseconds for HW
+		 * to recover... Let's try up to 10 times with 1ms delay
+		 */
+		usleep(1000);
+	}
+	return ok;
+}
+
+static int get_random_key(unsigned char *out_key)
+{
+	uint64_t val1, val2;
+
+	if (!get_rndr(&val1) || !get_rndr(&val2))
+		return 1;
+	memcpy(&out_key[0], &val1, sizeof(uint64_t));
+	memcpy(&out_key[8], &val2, sizeof(uint64_t));
+	return 0;
+}
+
+static int rekey(struct rng *ent_src)
+{
+	uint64_t thr;
+
+	message_entsrc(ent_src, LOG_DAEMON|LOG_DEBUG, "Rekeying...\n");
+
+	/* Grab new key & iv_buf from HW */
+	if (get_random_key(key) || get_random_key(iv_buf))
+		return 1;
+
+	/* Grab a new 8 bits random rekey threshold, we thus
+	 * rekey every 1 to 255 refills
+	 */
+	memcpy(&rekey_count, key, sizeof(rekey_count));
+	rekey_count &= ((1 << REKEY_BITS)-1);
+	return 0;
+}
+
+static int refill(struct rng *ent_src)
+{
+	message_entsrc(ent_src, LOG_DAEMON|LOG_DEBUG, "Refilling...\n");
+	if (--rekey_count < 0 && rekey(ent_src)) {
+		message_entsrc(ent_src,LOG_DAEMON|LOG_DEBUG,
+			       "failed to get AES seed randomness\n");
+		return 1;
+	}
+
+	/* Re-mangle the buffer */
+	aes_buf_pos = 0;
+	return ossl_aes_mangle(ossl_ctx, aes_buf, CHUNK_SIZE) < 0;
+}
+
+static int fill_from_aes(struct rng *ent_src, void *buf, size_t size)
+{
+	size_t chunk, avail, i;
+
+	for (i = 0; i < size; i += chunk) {
+		avail = CHUNK_SIZE - aes_buf_pos;
+		chunk = size <= avail ? size : avail;
+		memcpy(buf + i, &aes_buf[aes_buf_pos], chunk);
+		aes_buf_pos += chunk;
+		if (aes_buf_pos == CHUNK_SIZE && refill(ent_src))
+			return 1;
+	}
+	return 0;
+}
+
+static int init_openssl(struct rng *ent_src)
+{
+	ossl_ctx = ossl_aes_init(key, iv_buf);
+	if (ossl_ctx == NULL) {
+		message_entsrc(ent_src,LOG_DAEMON|LOG_DEBUG,
+			       "openssl initialization failed\n");
+		return 1;
+	}
+	return refill(ent_src);
+}
+
+static int fill_from_rndr(void *buf, size_t size)
+{
+	uint64_t r;
+	size_t chunk;
+
+	while (size) {
+		/* Grab 64-bits */
+		if (!get_rndr(&r))
+			return 1;
+
+		/* This is endian-broken but who cares for random
+		 * numbers ?
+		 */
+		chunk = size > 8 ? 8 : size;
+		memcpy(buf, &r, chunk);
+		size -= chunk;
+		buf += chunk;
+	}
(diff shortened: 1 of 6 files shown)


rngd_rndr.c
```diff
@@ -0,0 +1,105 @@
...
+			: "=r" (*rndr_val), "=r" (ok)
+			:
+			: "cc");
+		if (ok)
+			break;
+		/*
+		 * The spec implies we should wait some milliseconds for HW
+		 * to recover...
+		 */
+		usleep(2000);
```
nhorman, line 66: This suggests to me that you do want an aes mode of operation here, just because the combination of the loop and usleep could result in 20ms of latency per 64 bit ranndom value, it doesn't have to be on by default  but it would be nice to have

nhorman (changes requested):
Looks pretty good,  thx! One comment Inline about the lack of aes, but if you add that option in, this works for me

rngd_rndr.c
ozbenh: Hi ! Thanks for your reply.

I am not sure what benefit AES would bring here. RNDR is supposed to be pretty fast (at least on the implementations I'm aware of ;-) It's a DRBG that's only occasionally self-reseeded from a true entropy source, it can provides double digits MB/s.

From my understanding (again of existing implementations since the architecture is mostly useless in how it defines this), the failure case is a very rare case when the HW detected a fault and resets the unit, which can cause that few ms delay.

I can add AES as a disabled-by-default option if you wish, like I think Intel does, in fact Balbir original code that I took over did have it, but I didn't think it was particularly useful.

nhorman: It likely isn't useful in most cases, so I'm fine with disabling it by default, but I ran into the same issue with the power DARN instruction, in that someone tripped over a system that had a weaker implementation and wound up getting instruction failures in the common case, so I would feel better if users had a workaround in the field should they hit that condition on some loosely designed arm system.  Thanks!

ozbenh: Allright, will do !

ozbenh: Hrm... so I was getting bothered by the N copies of the openssl code that are almost-but-not-quite identical between the various sources and started looking at consolidating them in an ossl_helpers.c, when I noticed something weird:

Things like rngd_rdrand.c and rngd_jitter.c call an openssl_mangle() helper that uses "strlen(tmp)" on the "plaintext" source.

But that "plaintext" has no guarantee to have a terminating NUL character, does it ? It may also have randomly NUL bytes in there that are "normal" parts of the random numbers, or am I missing something ?

The rdrand code is hard to read and somewhat seems to be ok for buffer overflow because we only use AES_BLOCK * RDRAND_ROUNDS in the openssl case and the buffer is CHUNK_SIZE * RDRAND_ROUNDS, and since the buffer is static, it will be filled with zeros. But we might still hit a 0 in the actual random stream can't we ?

In the jitter case, it's trickier. As far as I can tell, we call xread_jitter for the whole size of the buffer, so unless I miss  something, we also have a potential issue.

I noticed that the "darn" version added an explicit size argument... What's your take on this ?

ozbenh: Correct me if I'm wrong here, but the destination buffer just needs to be the size of the input plus 16 (one AES block) to fit all combinations, no ?

ozbenh: I ended up hacking this: https://github.com/ozbenh/rng-tools/commit/28f779869d7673f1a698a43de51e0392abc87a11 let me know what you think

ozbenh: So now that I've done the above, what should I do with AES in your opinion ?

IE. one approach is the darn one: just re-encrypt the same buffer over and over and occasionally re-fill it from the HW. This means still having potential "pauses" during those refills if the HW isn't responsive. I'm not sure how much that would improve the overall situation.

Or I could do something where I fill a buffer with random from the HW, but if the HW fails, rather than waiting, I just use AES to encrypt the previous buffer as a source of extra "randomness" ?

What do you think ?

nhorman:
looks like you have a conflict in the branch above, caused by the fact that you had the ossl fixes in both this branch and the one I just merged.  if you can rebase and force push this branch, I'll gladly take it.  My expectation should be that this rebase will wind up reducing to the single rndr instruction commit

ozbenh:
Ah yes, I haven't had a chance to rebase fully yet (ran out of time yesterday). Any chance you can look at my question above ? Should I just copy what darn does ?

nhorman:
Yes apologies,I think the darn approach is sufficient - renencrypt the same
buffer with periodic rekeying from rndr.  We shouldn't use it in the
nominal case anyway, but I'm happy to accept either approach

ozbenh:
Would it make sense for the -t  option (test sink) to also run the fips tests on the generated random numbers ?

nhorman:
looks good, thanks!

nhorman (approved):

ozbenh:
Thanks !

Mingli-Yu:
@ozbenh 
But rndr failed to initialize as below in my env, any hints?

\# systemctl status rngd
* rngd.service - Hardware RNG Entropy Gatherer Daemon
     Loaded: loaded (/lib/systemd/system/rngd.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2021-07-05 07:15:47 UTC; 1min 31s ago
   Main PID: 152 (rngd)
      Tasks: 5 (limit: 2385)
     Memory: 1.8M
     CGroup: /system.slice/rngd.service
             `-152 /usr/sbin/rngd -f -r /dev/hwrng

Jul 05 07:15:47 qemuarm64 systemd[1]: Started Hardware RNG Entropy Gatherer Daemon.
Jul 05 07:15:48 qemuarm64 rngd[152]: Initializing available sources
Jul 05 07:15:48 qemuarm64 rngd[152]: [hwrng ]: Initialized
Jul 05 07:15:48 qemuarm64 rngd[152]: [rndr  ]: No HW SUPPORT
Jul 05 07:15:48 qemuarm64 rngd[152]: [rndr  ]: Initialization Failed
Jul 05 07:15:48 qemuarm64 rngd[152]: [jitter]: Initializing AES buffer
Jul 05 07:16:14 qemuarm64 rngd[152]: [jitter]: Enabling JITTER rng support
Jul 05 07:16:14 qemuarm64 rngd[152]: [jitter]: Initialized

nhorman:
Do you have a cpu that has rndr support?

Mingli-Yu:
@nhorman 
I run the system in a qemu env, how to check if the cpu support rndr or not?
root@qemuarm64:~# cat /proc/cpuinfo
processor	: 0
BogoMIPS	: 125.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 0

processor	: 1
BogoMIPS	: 125.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 0

processor	: 2
BogoMIPS	: 125.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 0

processor	: 3
BogoMIPS	: 125.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 0

root@qemuarm64:~#

nhorman:
you look at the output of rngd, from your log:
Jul 05 07:15:48 qemuarm64 rngd[152]: [rndr ]: No HW SUPPORT

I'm not sure what the sysfs cpu flag string translates into, but that message above is the result of rngd not finding support in the hw cap register, so the feature isn't supported

ozbenh:
Make sure you have a recent upstream qemu. If you are running qemu-system-img, that your kernel has the support as well

Mingli-Yu:
@nhorman 
Thanks for your update!

@ozbenh 
$ qemu-system-aarch64 --version
QEMU emulator version 6.0.0
Copyright (c) 2003-2021 Fabrice Bellard and the QEMU Project developers

The command which used to run qemu-system-aarch64 as below, BTW does rngd service works normally if I run into the error "Jul 05 07:15:48 qemuarm64 rngd[152]: [rndr ]: Initialization Failed"?
qemu-system-aarch64 -device virtio-net-device,netdev=net0,mac=52:54:00:12:34:02 -netdev tap,id=net0,ifname=tap0,script=no,downscript=no -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0 -drive id=disk0,file=/build/tmp-glibc/deploy/images/qemuarm64/wrlinux-image-std-qemuarm64.ext4,if=none,format=raw -device virtio-blk-device,drive=disk0 -device qemu-xhci -device usb-tablet -device usb-kbd  -machine virt -cpu cortex-a57 -smp 4 -m 2048  -m 2048 -serial mon:stdio -serial null -nographic -device VGA,edid=on -kernel /build/tmp-glibc/deploy/images/qemuarm64/Image--5.10.46+git0+b969f83647_139fe7d684-r0-qemuarm64-20210705070001.bin -append 'root=/dev/vda rw  mem=2048M ip=192.168.7.2::192.168.7.1:255.255.255.0 console=ttyAMA0 console=hvc0  selinux=0 enforcing=0'

ozbenh:
Replace -cpu cortex-a57 with -cpu max, I think it should help

Mingli-Yu:
@ozbenh 
Thanks very much! 
Updateing "-cpu cortex-a57" to "-cpu max" indeed silences the error, but I have no idea what the cpu model max stands for.

ozbenh:
It stands for "an imaginary CPU that has all the architecture features supported" :-) qemu doesn't (well didn't in 6.0.0, I haven't checked in a while) have a model for a "known" core with rng yet.

Mingli-Yu:
> It stands for "an imaginary CPU that has all the architecture features supported" :-) qemu doesn't (well didn't in 6.0.0, I haven't checked in a while) have a model for a "known" core with rng yet.

Does it mean there is no other cpu model except "max" support rndr? Or any hints to specify the cpu model parameters as
https://lists.openembedded.org/g/openembedded-core/message/153662?p=,,,20,0,0,0::Created,,qemuarm64.conf,20,2,0,84040384, ?

ozbenh:
Check the source :-)

Mingli-Yu:
> Check the source :-)

I don't quite understand what the source you mean.  Not only notice the message "[rndr ]: Initialization Failed" in qemu env, but also notice it on marvell-cn96xx board. Is it safe to ignore the message "[rndr ]: Initialization Failed"?

ozbenh:
I was going to say check the qemu source code to see if there's a way to enable the rng extension with another core model. That's what I would do. Yes it is safe to ignore that message.

Mingli-Yu:
@nhorman 
Considering it's safe to ignore the message "[rndr ]: Initialization Failed", is it possible to improve the message a little to something else as user may be afraid something is wrong when notify there is some message like "Initialization Failed"?

nhorman:
@Mingli-Yu like what?  The message is accurate (the entropy source failed initialization), and at some point you have to be able to use the other log messages and an understanding of how the software works to infer whats critical and whats not.

Mingli-Yu:
@nhorman  something like only print the message when the entropy source initialization succeeds. Though the current message is accurate (the entropy source failed initialization), but the user may think something is wrong if there is some message "[rndr ]: Initialization Failed" if they don't check the rng-tools source code.

nhorman:
Only print a message when something works? No.  What should I tell the first person who asks why an entropy source silently failed?  I'm sorry, but I'm not changing the message.  Its common for errors to not always be fatal, and its incumbent on the user to understand when an error is fatal and benign.

ozbenh:
To be completely fair, it might be valuable to make some of these a bit less scary :-) Maybe along the lines of "[rndr ]: Not available on this system". Oh well, we had various entropy sources failing for ... ever with similar scary messages and nobody complained so I don't know if it's really worth bothering

nhorman:
I liken entropy sources to web browser plugins (or any similar plugin architecture).  Sometimes they fail, and if they do they log a message to that effect, but the overall program (rngd in this case), keeps running, so the implication is that it must not be fatal.  We could change the error message, but its not possible in all (most) cases to determine if the failure was due to a malignant or benign reason.  The best we can do is report the cause as exactly as possible (in this case, due to lack of hardware availability).  I'll compromise with you though, and document in the man page the fact that individual entropy source failures are not fatal to operation.

PR #129, opened 2021-05-21
Create helpers for openssl AES use

ozbenh:
Followup on comments in https://github.com/nhorman/rng-tools/pull/128

This moves the openssl code that is duplicated in several sources
to some helpers. This also fixes the use of "strlen" on raw random
data which probably didn't do what someone thought it did ...

Note: rtlsdr could probably use the random key generation helper, but
for now I have, I think, kept the code mostly equivalent

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,7 +9,7 @@ man_MANS	 = rngd.8 rngtest.1
 noinst_LIBRARIES = librngd.a
 
 rngd_SOURCES	= rngd.h rngd.c	rngd_entsource.h rngd_entsource.c	\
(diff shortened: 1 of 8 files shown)


ozbenh:
BTW. I could only do rather cursory testing of this branch... I don't have all of the HW and I'm not entirely certain how you "verify" that the entropy generated is "good"... make test passes and rngd -t seems happy with jitter and rdrand but...

nhorman:
I just spent an hour piping rngd running only the jitter source (with AES enabled) through rngtest, and it passed the nist qualifications for the entire time, so I'm satisfied its working well

PR #138, opened 2021-07-11
Additions and fixes for the CPU detection code

nefigtut:
```
Per my understanding, these are important changes which
should be included to the `onecpu` branch before it gets
merged to the `master`. A build and tests are fine.

- Adjust the main check condition
- Adjust a type of a variable for the sysconf() result
- Add some code for errors and corner cases handling
- Adjust a return code so it matches a jitter library one
  (this is needed for some future improvement I'm thinking of)
- Add some spaces and new lines for a better looking code
```


nefigtut:
thanks a ton, Neil!

Issue #141, opened 2021-07-30
NULL-deref in an unlikely event of memory allocation failure


nefigtut:
Hello,

This code from `rngd_nistbeacon.c` will cause NULL-deref in an unlikely event of memory allocation failure:

```
static void get_json_byte_array(json_t *parent, char *key, char **val, uint32_t *len)
{
        ...
        newval = malloc(rawlen/2);
        ...
                        newval[j] = tmpval;

```

Unfortunately, I do not see how to fix this easily. In case we just add `if (!newval)` a NULL-deref  will happen later on accessing NULL pointed by `static struct nist_data_block block` `char *` fields.

nhorman:
ew, you're right.

This is the best way I can think to fix it:
https://github.com/nhorman/rng-tools/pull/142

Basically we add return codes to each type parser, and return 0 in the curl parse routine if any one fails.  That will cause curl to abort the whole parsing process and return an error to the caller of curl_easy_perform, which should fix the problem. The xread_nist function will fail, and the next call will attempt to update the nist block again.  Let me know what you think.

nefigtut:
> and the next call will attempt to update the nist block again. 

I do not quite understand the code, honestly. If memory starvation persists, will it make `rngd` to request a NIST block from a server again and again in an endless loop? An endless loop - because if a memory starvation happens, it usually persists and does not resolve by itself in a short time.

nhorman:
well, sort of.  If memory starvation persists, then yes, it will continue to try and fail to get the needed ram.  But that will only continue so long as the memory constraint exists.  Once that is relieved, we break out of the loop.  in the interim, other entropy sources will be tried to gain the needed entropy

nefigtut:
ok, got it. fair enough, i guess.

PR #142, opened 2021-07-30
Fix an ugly potential null deref

nhorman:
Vladis pointed out here:
https://github.com/nhorman/rng-tools/issues/141

That we had a potential null deref in nistbeacon, that wasn't easily
fixed, and thei is the best way I can come up with to address it

Basically, when we parse a nist json block returned from a server, we
need to be aware that some memory allocations that occur in the process
might fail. If they do, we should abort the curl_easy_perform operation
so that the calling fuction knows that the block is invalid. To do that
I have to check the return code of every element parser we have (which
we use several times).  Thats alot of extra if(rc){return 0} code, so
I'm codifying it in a variadic macro that returns if the call fails

yes, I'm returning in a macro. Yes, thats ugly.  And yes, there is a
huge comment warning all who enter that code that chaos and madness lie
within.

### DIFF
--- a/rngd_nistbeacon.c
+++ b/rngd_nistbeacon.c
@@ -285,12 +285,17 @@ static int get_json_byte_array(json_t *parent, char *key, char **val, uint32_t *
  * do that we need to return 0 from parse_nist_data_block, so the curl library
  * knows to abort the operation and fail in the call to curl_easy_perform.  That
  * in turn means checking a ton of return code when parsing out individual
- * elements.  To do that, we codify the individual element parse call, rc chedk,
+ * elements.  To do that, we codify the individual element parse call, rc check,
  * and return here in this macro.  Yes, it means returning from a function in a
  * macro, which is ugly, but thats why I'm writing this huge comment here, so
  * you won't be caught off guard.  You've been warned.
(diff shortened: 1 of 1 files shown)


nefigtut:
Honestly, I do not see how this can be handled with less of a boilerplate code. So I'm voting for this solution, with a #define, etc.

A couple of notes, if I may.

Would it be possible to add some messaging for the out-of-memory case? The same way it is done by `message(LOG_DAEMON|LOG_ERR, "Byte array isn't of even length!\n");` just couple of lines above? I believe, a user or a troubleshooting admin would want to know why the NIST block was invalidated.

There is just a mistype in a long comment: "rc chedk", an easy fix, I guess.

But in general I believe this fix is fine. Thank you, Neil for this!

nefigtut:
hello, Neil,
it looks like there is another small issue with logging, do you want me to post an MR? or can you please do a fix, i believe, it is just faster. my point is that we need to stringify the call argument:

```
message_entsrc(ent_src, LOG_DAEMON|LOG_ERR, "Out of memory in %s\n", #call );
```

with that, i believe the code is ready to be merged.

PR #149, opened 2021-08-28
Block termination signals for the jitter tasks

frankencode:
To improve reliability of daemon shutdown by making sure the foreground
thread is the thread catching SIGINT and friends.

Explanation:
I'm still experiencing random freeze ups with rngd (>=5.9) on one of my embedded targets.
(Adding the full 90s systemd default timeout to a reboot.)

Disabling the jitter entropy source (-x jitter) seemed to solve the freeze ups.
Analyzing the problem I found it difficult to figure out what happens in which order
because I cannot say which of the running threads actually catches the SIGINT
and hence I decided to block SIGINT in the jitter tasks and thereby making sure
the shutdown magic starts in a more predictable manner.

### DIFF
--- a/rngd_jitter.c
+++ b/rngd_jitter.c
@@ -257,6 +257,7 @@ void jitter_thread_exit_signal(int signum)
 
 static void *thread_entropy_task(void *data)
 {
+	sigset_t blockset;
 	cpu_set_t cpuset;
 
 	ssize_t ret;
@@ -265,6 +266,12 @@ static void *thread_entropy_task(void *data)
 	struct timespec start, end;
 	int written;
 	/* STARTUP */
+
+	sigemptyset(&blockset);
+	sigaddset(&blockset, SIGINT);
+	sigaddset(&blockset, SIGTERM);
+	sigaddset(&blockset, SIGALRM);
+	pthread_sigmask(SIG_BLOCK, &blockset, NULL);
 
 	/*
 	 * Set our timeout value


nhorman:
Ok, theres nothing expressly wrong with what you've done, but you're explanation for why it helps anything doesn't make any sense either, and so I'm pretty hesitant to take this change.

In answer to your question about which thread catches sigint, the answer is "one of them".  its indeterminate at build time which thread might handle that signal, but it should also be completely irrelevant, because all it does is set the server_running flag to false, so that the main thread breaks out of its forever loop in the do_loop function.  The only thing that should be able to happen in the jitter loop is that we have a thread return from write early, with errno == EINTR, and we ignore that.  Best thing to do would be build  from the head of the tree, run the daemon until the problem occurs, then attach gdb to the process, and dump a backtrace of all the threads.  If you post those here, I can tell you whats going on.

frankencode:
Hi Neil,

thanks for your quick reply. I don't think EINTR is always handled properly.

For instance in thread_entropy_task(), rngd_jitter.c:237ff 
```c++
		while(written != me->buf_sz) {
			message_entsrc(me->ent_src,LOG_DAEMON|LOG_DEBUG, "Writing to pipe\n");
			ret = write(me->pipe_fd, &tmpbuf[written], me->buf_sz - written);
                        if ((ret < 0) && (errno != EBADF))
				message_entsrc(me->ent_src,LOG_DAEMON|LOG_WARNING, "Error on pipe write: %s\n", strerror(errno));
			message_entsrc(me->ent_src,LOG_DAEMON|LOG_DEBUG, "DONE Writing to pipe with return %ld\n", ret);
			written += ret;
               }
```
What happens when ret = -1 and errno = EINTR ? Or shouldn't I be to much concerned because it is non-blocking?

nhorman:
write is blocking, but if this thread handles SIGINT, write will return with errno set to -EINTER, and ret = -1, which should cause the loop to simply continue.  Unless you have some evidence to the contrarary?

frankencode:
Sure, and ``written`` will be decremented because ``ret`` requals -1.

frankencode:
And if ``written`` turns negative the next ``write`` would write ``me->buf_sz - written`` bytes, which will lead to an undefined memory read beyond the end of the buffer and therefore possibly lead to a crash.

nhorman:
Sure, thats possible, fix would be easy, just check for EINTR, and continue the loop if thats the errno value. 

Are you seeing a crash though?

frankencode:
Yes, the fix is easy in this instance. In general these concerns can be avoided by setting up proper signal masks.

nhorman:
Sure, and I'm not opposed to doing it on principle.  What I am opposed to doing is accepting it as a fix for the hang you encountered, when I can't see an clear correlation between the behavior you initially described (a hang), and the fix you propose (blocking the SIGINT signal for the jitter threads).  

Can you provide a backtrace of all the threads when you encountered this hang?  That would help identify the problem (and hopefully show a correlation to signal handling)

frankencode:
A quick code review shows me your signal handling code is either missing or flawed. Reducing the ways in which SIGINT is handled seemed prudent to me, when I see a hangup right after delivery of SIGINT. (Btw. systemd sends SIGINT followed by SIGCONT). Attaching a gdb is a bit difficult for me, because it hangs every second day and always during system shutdown/reboot, when I have already lost my terminal.  You are of course free to edit or not take any patch. I already learned something from your replies. Thanks for that!

nhorman:
I have a theory, and it involves signal handling, but not in the way you are thinking.  I'm thinking that, because I send SIGUSR1 to a thread before I call pthread_mutex_lock/pthread_cond wait, we get a race in which the entropy thread exits prior to our wait, and so we block forever waiting for the completion .  Can you augment your rngd service unit file to enable debug logging (i.e. add the -d flag to the ExecStart line).  Next time the hang occurs upload the logs for the unit here please.  I have a feeling we are going to see close_jitter_entropy_source indicate that its waiting for that condition that never arrives.

frankencode:
Good idea. I've added the "-d" and I'm keeping my screen log running. I'll post the log once I notice another hang up...

nhorman:
Thank you, if this is what I think it is, I can have a patch for you ready same day that you report the logs

nhorman:
thats a bit strange.  I was expecting to see one extra debug messages there, that I don't currently see.  But it turns out the patch for my hypothesis is 2 lines long, so lets give it a shot anyway.

https://github.com/nhorman/rng-tools/tree/shutdown-race

Has the changes to avoid the shutdown race I think might be a problem if you want to give it a shot

frankencode:
So I had finally time to look at this issue again. Your 2 line patch makes absolutely sense. Definitely a deadlock condition there fixed. The signalling of `statecond` could have happened before the parent thread starts waiting for it.

Though, I don't trust the shutdown logic much. The Linux man on setjmp(3) for instance states:

> The only way of avoiding undefined behavior is to  ensure one of the following:
>       *  After long jumping from the signal handler, the program does not call any non-async-signal-safe functions...

See: https://man7.org/linux/man-pages/man3/longjmp.3.html

And the first thing you do after the siglongjmp in rngd_jitter.c:
```C++
out_interrupt:
	free(tmpbuf);
```
But free(3) is not async signal safe on my customer's platform.
So in theory the rest of the program's shutdown procedure is undefined anyways.

nhorman:
I'll check in the fix shortly.

As for your new concern, I'm not sure I see the problem.  I agree that siglongjmp is not async-signal-safe, but as I understand it, the safety issue there is that non-reentrant functions (like free), might operate on inconsistent data (e.g. if someone frees a buffer and gets a signal during the process that also frees the buffer, one of those two contexts will operate on an invalid pointer).  That isn't the case here though.  The tmpbuf that you are referring to is per-thread storage, and each thread is guaranteed to be the only allocator/free-or of that pointer.  given that we use pthread_kill to issue the SIGUSR1 signal to each jitter thread, and then call pthread_join to wait for the exit of that thread, I don't see a way to trigger a duplicate free condition such as the one you are alluding to.

Well, thats not 100% true.  I could see a way that a thread could intercept a second SIGUSR1, while in the middle of the free operation,  jump back to the sigsetjmp location and attempt to preform a second free operation.  But such an operation would:

1) need to be triggered with external stimulus (i.e. since the main thread is blocked in pthread_join, we would need an external program sending SIGUSR1 to the rngd process frequently, in the hopes that the 'right' thread intercepts it

2) result not in a hang (as I assume you are suggesting), but rather an invalid free that (IIRC should cause SIGABRT/core dump)

Thats not to say we can't improve the code here, but if you think such an operation is causing an issue in your environment, I think we would need some more evidence to support that before fully understanding and implementing a fix for it.

frankencode:
free(3) operates in many cases on complex caching queues for freed memory chunks and also on free chunk lists. Those double linked list operations you do not want to interrupt and call again in the interrupt handler. Dangling free lists then are used in the next malloc/free() leading to funny results... If your jitter thread catches its SIGUSR1 in a malloc() or free() your signal handler runs into undefined behavior if it calls free().
Anyways, I'm porting back your two-liner patch for my customer. Thanks for your help!
(And yes you do not need to merge this pull request. The discussion went midely off-topic;)

PR #154, opened 2021-12-14
Add armv6l to list of pkcs11 detected arches

nhorman:
This will properly set the PCKS11 library search path for armv6l devices

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -124,7 +124,7 @@ AS_IF(
 		PKG_CHECK_MODULES([libcrypto], [libcrypto], [], [AC_MSG_ERROR([libcrypto is required])])
 		AC_DEFINE([HAVE_PKCS11],1,[Enable PKCS11])
 		AS_CASE([$host_cpu],
-			[i?86|armv7hl],
+			[i?86|armv7hl|armv6l],
 			[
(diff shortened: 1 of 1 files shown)


ptiggerdine (commented):
aside from change /usr/lib to /usr/lib/arm-linux-gnueabihf/ it worked a treat

nhorman:
oh, if you are cross compiling and want to set your path to something non-standard, you can do that post configure time, at build time, like this:
make PKCS11_ENGINE=/usr/lib/arm-linux-gnueabihf/lib

ptiggerdine:
Just to be clear, I'm not cross compiling. Just appears to be where the distro as put the libs.

nhorman:
ok, thats fine, but regardless, /usr/lib/arm-linux-gnueabihf isn't a standard location (i.e. not /usr/lib or /usr/lib64), and to avoid mistakes in automatically searching for libs (like in the case where multiple arches/versions of the library may exist), the proscribed solution is to specify the library location at make time, as above

ptiggerdine:
Thanks for your help and fast assistance with this. I was going to create the PR myself, just beat me to it :)

PR #158, opened 2022-02-09
Fix up version option handling

nhorman:
The man page was out of date, specifying -V as the man page version
flag, when It should be -v.  Also, the arg parser wasn't handling those
options well, as it was returning an error for those valid options.  Fix
that up to just exit cleanly after printing the version info

### DIFF
--- a/rngd.8.in
+++ b/rngd.8.in
@@ -24,7 +24,7 @@ rngd \- Check and feed random data from hardware device to kernel random device
 [\fB\-R\fR, \fB\-\-force-reseed=\fInnn\fR]
 [\fB\-q\fR, \fB\-\-quiet\fR]
 [\fB\-?\fR, \fB\-\-help\fR]
-[\fB\-V\fR, \fB\-\-version\fR]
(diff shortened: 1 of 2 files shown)


c0bw3b:
I built v1.65 with this patch on top

```session
$ rngd -v ; echo $?
rngd 6.15
Copyright 2001-2004 Jeff Garzik
Copyright 2017 Neil Horman
Copyright (c) 2001 by Philipp Rumpf
This is free software; see the source for copying conditions.  There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
0

$ rngd -V ; echo $?
rngd: invalid option -- 'V'
Try `rngd --help' or `rngd --usage' for more information.
64

$ ./result/bin/rngd --version 2> /dev/null ; echo $?
0
```

The exit codes are consistent now.

Even though the exit code is 0 the version and copyright message outputs on **stderr** when calling `rngd --version` or `rngd -v`.
I can't really say if that is the expected behaviour.
`rngtest --version` will output to stdout and return 0 for example.

PR #168, opened 2022-04-15
Build Power9 code separately

nefigtut:
The PPC64 DARN instruction is defined in ISA 3.0 which is for Power9 CPUs.
Power8 CPUs are ISA 2.07 and thus do not support DARN. Recently gcc has
limited the DARN instruction to power9 target only. Build a code with DARN
in a separate library with special compile flags.

Also add a missing check for get_darn() return value as in other places.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,10 +19,6 @@ if RDRAND
 rngd_SOURCES	+= rngd_rdrand.c rdrand_asm.S
 endif
 
-if DARN
-rngd_SOURCES	+= rngd_darn.c
-endif
-
 if RNDR
 rngd_SOURCES	+= rngd_rndr.c
 endif
@@ -36,6 +32,17 @@ rngd_SOURCES	+= rngd_rtlsdr.c
 endif
 
 rngd_LDADD	= librngd.a $(LIBS) $(librtlsdr_LIBS) ${libp11_LIBS} ${libcrypto_LIBS} ${jansson_LIBS} ${libcurl_LIBS} ${libxml2_LIBS} ${openssl_LIBS} ${libcap_LIBS} $(PTHREAD_LIBS)
+
+if DARN
+rngd_SOURCES	+= rngd_darn.c
+
+# Build Power9-only code with Power9
+# compile flags in a separate library
+rngd_LDADD		+= libdarn_impl.a
+noinst_LIBRARIES	+= libdarn_impl.a
+libdarn_impl_a_SOURCES	= darn_impl.c
+libdarn_impl_a_CFLAGS	= -mcpu=power9 -mtune=power9
+endif
 
 if PKCS11
 rngd_SOURCES	+= rngd_pkcs11.c
--- a/darn_impl.c
+++ b/darn_impl.c
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2017, Neil Horman
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms and conditions of the GNU General Public License,
+ * version 2, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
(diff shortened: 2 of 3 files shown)


dermotbradley:
I assume related to this, rng-tools recently failed to rebuild for Alpine on PPC with a DARN-related issue.

I fixed it using this patch to binutils:

https://git.alpinelinux.org/aports/tree/main/binutils/binutils-ppc-fix-machine-options.patch

nefigtut:
hi, @dermotbradley, i'm not sure i understand properly. DARN was introduced in ISA3.0 (power9), so gcc/gasm refusing to recognize DARN with power8 target does seemingly a right thing.

it is not that i understand your patch in full. do you mean that this rng-tools commit can be reverted after your patch is applied to gas and the update reaches a build machine? does this imply that DARN will be recognized even with `gcc -mcpu=power8`?

dermotbradley:
@nefigtut All I know if that with the newer binutils (2.38) installed the Alpine Linux CI (running on a Power8 machine) suddenly started to fail building the ppc64le package of rng-tools (it was "gas" rather than "gcc" failing) whereas it was previously able to do so with binutils 2.37. When the patch I linked to above was applied to binutils 2.38 then the CI was able to build ring-tools fine.

The DARN instructions are indeed Power9 instructions but this should not prevent a Power8 machine from building code using them - these instructions should only affect **runtime**.

dermotbradley:
> gcc -m64 -mcpu=power8

That is compiling code for be run on a Power8 machine so obviously that code cannot contain a Power9 instruction. In the same way that if you were to compile code for a 486 (x86) then it could not contain things like SSE3/AVX2 instructions. From a quick "grep" of the rng-tools source code I don't see that "-mcpu=power8" defined anywhere so I guess its coming from your system/environment.

I patch I mentioned earlier that I used on Alpine is not my patch, it was one someone else had already submitted to binutils: https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=cebc89b9328;hp=9cbed90ee623d75e31994e7943960c997ba565f2


Actually I'd expected the DARN-related code in rngd to act like the RDRAND code for x86/x86_64 and the RNDR code for arm to check at run-time if the particular CPU can handle the relevant instruction like:

https://github.com/nhorman/rng-tools/blob/master/rngd_rndr.c#L185
https://github.com/nhorman/rng-tools/blob/master/rngd_rdrand.c#L257

nefigtut:
> I don't see that "-mcpu=power8" defined anywhere so I guess its coming from your system/environment.

yes, this comes from fedora's build system and this initially has created this issue. with that, we should explicitly build a piece of code with `DARN` with `-mcpu=power9`. gcc (current version in a current fedora-latest, at least) also refuses to build a DARN code without `-mcpu` and requires `-mcpu=power9` explicitly. this means power8 is a default gcc target on a ppc64.

>  to check at run-time if the particular CPU can handle the relevant instruction

exactly. rng-tools does this in [rng-tools/blob/172bf0e3dad63cfc3959dd69941d64d9a978cb6b/rngd_darn.c#L183](https://github.com/nhorman/rng-tools/blob/172bf0e3dad63cfc3959dd69941d64d9a978cb6b/rngd_darn.c#L183).

> I patch I mentioned earlier that I used on Alpine is not my patch, it was one someone else had already submitted to binutils

i do not understand what this patch does. my point is: if this patch allows a `DARN` code to be built for power8 gcc target (which is default gcc ppc64 target when no `-mcpu` is given, as of now) - i do not think this is correct. as you mention above - "the same way ... code for a 486 (x86) ... could not contain things like SSE3/AVX2 instructions".

nhorman:
@dermotbradley the patch is fine.

What @nefigtut did was isolate the darn instruction to its own c file/library so that it could be build with -mcpu power9.  As you note, that's suspect because if you build that code for power9 and run on a power8 system, you'll get a SIGILL.  But we will never run that code on a power8 system, because we have a runtime check in init_darn_rngd (or whatever the darn entropy source init function is called), which checks the CPU support registers.  If the CPU at runtime doesn't support DARN, then the entropy source fails initialization, preventing the darn instruction from ever running.

This whole patch was done solely to allow build farm that compiles targeting power8 to support being able to run on power9 systems with the darn instruction enabled

dermotbradley:
@nhorman Ok, I'll try building on Alpine on ppc64le with this patch applied and see if it completes.

I don't use ppc64le myself - on Alpine packages are, by default, built for all supported arches - and so I won't have a way to test that it actually works as intended.

PR #175, opened 2022-06-29
extend rngtestjitter timeout to a 30 second

sharkcz:
Seems the current 15 second timeout is causing occasional false failures
after recent changes.


nhorman:
seems sensible, since jitter can be highly latent on some systems

Issue #174, opened 2022-06-28
rngtestjitter.sh failures after recent changes


sharkcz:
I am seeing rngtestjitter.sh failures on various platforms in our CI after recent changes (last batch of commits). 
````
[jenkins@ibm-p8-generic-02-guest02 ppc64le]$ cat tests/test-suite.log
==========================================
   rng-tools 6.15: tests/test-suite.log
==========================================

# TOTAL: 3
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: rngtestjitter.sh
======================

Disabling 0: Hardware RNG Device (hwrng)
Disabling 2: Intel RDRAND Instruction RNG (rdrand)
Disabling 1: TPM RNG Device (tpm)
Initializing available sources
[darn  ]: No HW SUPPORT
[darn  ]: Initialization Failed
[jitter]: Initializing AES buffer
[jitter]: Enabling JITTER rng support
[jitter]: Initialized
[pkcs11]: No pkcs11 slots available
[pkcs11]: Initialization Failed
[rtlsdr]: Initialization Failed
unable to adjust write_wakeup_threshold: Permission denied
rngtest: bits received from input: 2000064
rngtest: bits sent to output: 1980000
rngtest: FIPS 140-2 successes: 99
rngtest: FIPS 140-2 failures: 1
rngtest: FIPS 140-2(2001-10-10) Monobit: 0
rngtest: FIPS 140-2(2001-10-10) Poker: 0
rngtest: FIPS 140-2(2001-10-10) Runs: 1
rngtest: FIPS 140-2(2001-10-10) Long run: 0
rngtest: FIPS 140-2(2001-10-10) Continuous run: 0
rngtest: input channel speed: (min=847.269; avg=78187.550; max=19531250.000)Kibits/s
rngtest: FIPS tests speed: (min=126.314; avg=128.632; max=129.752)Mibits/s
rngtest: output channel speed: (min=1.552; avg=23.342; max=18.626)Gibits/s
rngtest: Program run time: 11832646 microseconds
killing
FAIL rngtestjitter.sh (exit status: 1)
````
Looks to me the 15 sec timeout needs to be extended. 30 secs seems to be a good value, the test passes consistently (on a Power8 system).

But on a aarch64 system (APM Mustang) I still seeing failures
````
[...]

PR #178, opened 2022-07-18
rngd_jitter: make timeout configurable

sharkcz:
The current default 5 sec timeout can be too short in some situations
(eg. slow hardware). Introduce an enviroment variable
RNGD_JITTER_TIMEOUT to make the value configurable. Increasing the
default timeout is not desirable, because it might slow down system
startup when there are other entropy sources.


nhorman:
why an environment variable instead of an option?  An option would be in keeping with how we handle entropy specific behavior.  See JITTER_OPT_THREADS and friends in rngd.c for examples.

sharkcz:
> why an environment variable instead of an option? An option would be in keeping with how we handle entropy specific behavior. See JITTER_OPT_THREADS and friends in rngd.c for examples.

because that was the first idea (https://github.com/nhorman/rng-tools/issues/174#issuecomment-1171401748) how to allow changing the timeout with minimal (or zero) changes in other code to get the tests passing. But you are right, an option would work too, together with a slight update to the test script.

sharkcz:
Reworked to use command line option instead.

Do you want the test change be a separate commit?

nhorman:
nah, single commit is fine @sharkcz .  This looks much better, thanks!

Issue #164, opened 2022-03-25
rng-tools stopped to build with jitterentropy v3.4.0


nefigtut:
Now a build errors out with:

```
rngd_jitter.c: In function 'rngd_notime_start':
rngd_jitter.c:56:45: error: invalid use of undefined type 'struct jent_notime_ctx'
   56 |         ret = -pthread_attr_init(&thread_ctx->notime_pthread_attr);
```

It looks like this is result of a commit https://github.com/smuellerDD/jitterentropy-library/commit/8373b05f1a55878f348d365f05b046ff42d2f9ce which now hides `struct jent_notime_ctx` under `#ifdef JENT_CONF_ENABLE_INTERNAL_TIMER` in `jitterentropy.h`.

jitterentropy's `Makefile` sets `JENT_CONF_ENABLE_INTERNAL_TIMER` unconditionally:

```
# Enable internal timer support
CFLAGS += -DJENT_CONF_ENABLE_INTERNAL_TIMER
```

but rng-tools which use `jitterentropy.h` do not touch `JENT_CONF_ENABLE_INTERNAL_TIMER` at all, so `struct jent_notime_ctx` becomes hidden.

It looks to me **(a personal opinion which can be wrong)** that this is a design error in the jitterentropy library itself. Users of a library may not know how the library was built. I.e. I believe, `#ifdef JENT_CONF_ENABLE_INTERNAL_TIMER` should not be used in `jitterentropy.h`.

Unfortunately, I was not able to invent a good solution. I've come up with a dirty hack (which is wrong and should not be like this, but it works) of defining `#ifdef JENT_CONF_ENABLE_INTERNAL_TIMER` in rng-tools' `configure.ac`. The diff is:

```
--- configure.ac	2022-03-24 13:14:11.000000000 +0100
+++ configure.ac.new	2022-03-24 15:58:56.187367770 +0100
@@ -95,7 +95,10 @@ AS_IF(
 				[AM_CONDITIONAL([JITTER], [true])
 				AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])
 				AC_CHECK_LIB(jitterentropy, jent_notime_settick,
-				[AC_DEFINE([HAVE_JITTER_NOTIME],1,[Enable JITTER_NOTIME])],
+				[
+					AC_DEFINE([HAVE_JITTER_NOTIME],1,[Enable JITTER_NOTIME])
[...]

PR #180, opened 2022-09-21
configure.ac: fix jitterentropy internal timers detection

nefigtut:
Use visibility of jent_notime_settick() to detect if jitterentropy library is built with internal timers. The code related is:

```
[ src/jitterentropy-timer.h ]
#ifdef JENT_CONF_ENABLE_INTERNAL_TIMER
int jent_notime_settick(struct rand_data *ec);
#else /* JENT_CONF_ENABLE_INTERNAL_TIMER */
static inline int jent_notime_settick(struct rand_data *ec)
{ (void)ec; return 0; }
#endif /* JENT_CONF_ENABLE_INTERNAL_TIMER */
```

Fix couple of whitespace and unused variable issues to please code scanners.

Thanks Neil for qcrypt, this looks like a great feature!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,9 +104,12 @@ AS_IF(
 		AC_SEARCH_LIBS(jent_version,jitterentropy,
 				[AM_CONDITIONAL([JITTER], [true])
 				AC_DEFINE([HAVE_JITTER],1,[Enable JITTER])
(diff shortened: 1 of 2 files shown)


nefigtut:
this would fix https://github.com/nhorman/rng-tools/issues/164

PR #186, opened 2022-12-01
rngd_jitter: decreased a thread on some of ARM platform

chenxy1988:
Dear maintainer,

Due to ARM cpu pipeline less than x86 or other arch, it would use more effort to generate data from jitter. this might block other process during system bootup, so decreased a thread to reserve resource for other process when system bootup. Just tested on armv7 and aarch64 platform only, so add armv7 & aarch64 macro only.

Thanks!

### DIFF
--- a/rngd_jitter.c
+++ b/rngd_jitter.c
@@ -454,7 +454,11 @@ int init_jitter_entropy_source(struct rng *ent_src)
 		CPU_SET(0, cpus);
 	}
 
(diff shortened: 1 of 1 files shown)


nhorman:
why would you do this? Given that:
1) Theres no evidence that reducing thread count by 1 unilaterally improves performance on all arm cpus
2) Theres already an administrative command line switch that allows you to specify the number of threads that are started for the jitter source

I don't see why we need arch specific controls as well.  It just makes things more complicated

Issue #197, opened 2023-05-18
Socket interface for hardware TRNGs


g-v-egidy:
Consider you have a hardware TRNG (for example based on zener noise, radio noise or similar) and want to feed it's data into the kernel. Then you need some hardware specific code that reads in the random data for example over serial or usb, maybe postprocesses and checks it, and then feeds it into the kernel. You probably want to mix this data with the other random sources available to rngd. For the latter I think you'd need an interface into rngd.

I think it would be best not to clutter rngd with drivers for many different hardware rngs. Because it limits the way the drivers can be written and makes it more complicated to distribute new drivers. Think programming languages, build system, dependencies, long-term-stable distributions having old versions of rngd, license.

How about offering a easy to use interface for this instead? I'm thinking about using a unix socket:
- rngd would get a "socket" driver
- parameters for it would be the path, file access mode, user, group.
- rngd creates the socket and waits for the external driver to push data into it
- the socket would be blocking: rngd only reads data from the socket when it needs new data. The hardware specific driver would push data into it until the socket blocks. 
- there wouldn't be a special protocol used for this, raw random bytes are written to the socket. rngd mixes them with the other random sources and fills them into the kernel entropy pool
- would also allow the hardware rng to be somewhere remote. Then something like socat could be used to forward the entropy over a secured network channel.

I think the alternative would be to completely leave rngd out of the picture and directly fill data from the hardware rng driver into the kernel entropy pool. rngd would do the same and the entropy from both is mixed by the kernel. When using a modern kernel and the `--force-reseed=nnn` option, the hardware rng driver and rngd both would do a force reseed every nnn seconds.

Does the socket driver idea make sense to you?

Which alternative do you think is the better approach, especially in the light of new kernels without entropy fill watermark?

nhorman:
What you are describing is the purpose of rngd.  Its interface to the kernel is the ioctl interface of /dev/random (or whatever character device you have that leads to the random pool).  The kernel takes any entropy received from that interface and mixes it with existing entropy.

rngds purpose is to gather that new entropy from various sources, for which there are rngd drivers (or entropy soruces in rngd parlance)

This is done purposefully for two reasons:
1) Adding new kernel drivers to collect entropy is significatnly more effort than adding collection code to a user space application
2) The selection of which entropy sources to draw from is generally considered an administrative decision rather than kernel based policy

[...]

PR #199, opened 2023-11-05
add "namedpipe" entropy source

g-v-egidy:
Reads entropy from a named pipe (aka FIFO). Another program, for example a driver reading and preparing data from an external hardware RNG, is expected to feed entropy into this named pipe. Should the named pipe be closed, rngd will try to reopen it.

This is option 1 as discussed in #197.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,7 +9,8 @@ man_MANS	 = rngd.8 rngtest.1
 noinst_LIBRARIES = librngd.a
 
 rngd_SOURCES	= rngd.h rngd.c	rngd_entsource.h rngd_entsource.c	\
-		  rngd_linux.h rngd_linux.c util.c ossl_helpers.c
+		  rngd_linux.h rngd_linux.c util.c ossl_helpers.c	\
+		  rngd_namedpipe.c
 
 if NISTBEACON
 rngd_SOURCES	+= rngd_nistbeacon.c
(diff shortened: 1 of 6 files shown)


nhorman:
I don't have any issue with this per-se, but how is it different from setting -r /path/to/pipe

g-v-egidy:
using the namedpipe option has several advantages over reusing `-r` for ingesting data from a hwrng driver daemon:
- it allows you to add a random source in addition to `/dev/hwrng`. So when your mainboard chipset has a supported hwrng it will show up as `/dev/hwrng`. When you now want to add an additional external hwrng to mix random data from several sources, you can use the namedpipe option to do that. if you used `-r` instead of namedpipe, you'd loose the random data from the mainboard chipset because `-r` can only be used once and can only get data from one source.
- namedpipe as implemented in this PR has automatic handling and reopening for EOF. So if you use a userspace daemon to talk to your hwrng and have it pipe random data to rngd, you can restart the userspace daemon and it will be able to continue to feed data to rngd afterwards. When using `-r` and the other side closes the pipe, rngd won't be able to continue to read from a new remote side afterwards and you'd have to restart rngd also.
- namedpipe has a configurable timeout

The timeout and EOF thing aren't really an issue when you use the default `-r /dev/hwrng` from the kernel, but they become an issue when you use a userspace daemon to talk to your hwrng-device. namedpipe is better optimized for that case.

nhorman:
yeah, ok, that seems reasonable, thank you!

g-v-egidy:
thanks for reviewing & merging.

PR #200, opened 2023-11-05
improve debug output to better show FIPS failures and source selection

g-v-egidy:


nhorman:
patch looks good, in the future if you could provide a description in the PR for record keeping, I would appreciate it.

PR #202, opened 2023-12-17
add option "force_as_fast_source" to jitter source

g-v-egidy:
By default the jitter entropy source is considered a "slow" source, meaning it's output is only used when no "fast" sources are available or able to provide sufficient random data. When setting this option to 1 jitter will be considered as a "fast" source too and it's output will always be considered (default 0).

Having this option is useful when you want to feed the kernel data from a mix of as many independent sources as possible. The concern of jitter being too slow to supply a sufficient amount of entropy data is only relevant when rngd needs to supply entropy to the kernel often and in short order. With modern kernels not signaling a low watermark anymore and thus only the known force-reseed time is used, you can ensure that jitter will be able to provide enough data at configuration time.

When comparing the default timeout of 5 seconds for the jitter source with the default force-reseed time of 5 minutes, you can see that jitter will often be able to provide enough data.


nhorman:
This change has the effect of making what is generally considered a 'slow' source a 'fast' source, which has other side effects.  Most notably, if you mark jitter as a 'fast' source, and it doesn't produce entropy in 100 rounds of collection, it will be disabled in perpetuity.

It might be better, instead of creating an option specifically to consider jitter to be 'fast', rather to create a global option to consider slow sources always (i.e. something like --use-slow-sources), which could be tested for in do loop, unilaterally setting the try_slow_sources variable on each iteration.

g-v-egidy:
Yes, a `--use-slow-sources` option would also work and not have the downside of the sources becoming disabled after too many failures.

I will create another pull request for that this evening.

Issue #201, opened 2023-11-05
Coarse mixing of multiple entropy sources


g-v-egidy:
when rngd currently has multiple entropy sources available, the way they are used is like this (pseudocode):
```
FIPS_RNG_BUFFER_SIZE=2500
random_step=64

for src in sources:
   entropy = read_from_source(src, FIPS_RNG_BUFFER_SIZE)
   if (fips_test_ok(entropy)):
      for(p=0; p<FIPS_RNG_BUFFER_SIZE; p+=FIPS_RNG_BUFFER_SIZE):
          feed_entropy_to_kernel(entropy[p], random_step)
```

A large chunk of bytes is read from a source, tested, and then fed in by default 39 steps to the kernel until finished. Only then the next entropy source is considered. When considering a modern kernel without observable fill watermark (>= 5.18), and using --force-reseed with the default option of 5 minutes, this means entropy from one source is fed to the kernel over the course of 195 minutes before the next source is used. This leads to quite a coarse mixing of different entropy sources.

It is my understanding that mixing entropy from different sources at least helps should one of the sources be not as random as expected.

The coarse way the mixing is done right now doesn't really facilitate this.

I suggest to change the way the mixing is done to something like this:
- first get a block with FIPS_RNG_BUFFER_SIZE of entropy from each enabled source
- check the blocks, only take the ones that pass the fips test
- loop over the blocks we got, taking one byte from each and feed the result to the kernel

This will lead to a byte-wise mixing of the different entropy sources.

Does this make sense?

nhorman:
Seems reasonable, would you like to write a patch?

g-v-egidy:
> would you like to write a patch?

yes, I can put developing it on my todo. But it will probably take a few days until I get to it.

Could you take a look at my open pull requests meanwhile?

nhorman:
ack, mered/notes made

PR #203, opened 2023-12-17
mix data fed to the kernel byte-wise from multiple sources

g-v-egidy:
Before this patch entropy data was gathered from one source, fed to the kernel, then the next source was gathered, fed, and so on.

With the default config on a modern kernel this lead to one source being used for 195 minutes before switching to the next source. Should one source not be as random as expected, the impact of this problem can be reduced by mixing the bad entropy with data from other sources. The long usage time of 195 minutes for one source didn't facilitate this well.

This patch changes this behavior and first gathers data from each enabled source and checks it. then a byte from each valid source block is taken in turn until the block to feed to the kernel (random_step) is filled. This mixed data is then fed to the kernel.

This fixes issue #201

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -759,46 +759,77 @@ static error_t parse_opt (int key, char *arg, struct argp_state *state)
 static struct argp argp = { options, parse_opt, NULL, doc };
 
 
-static int update_kernel_random(struct rng *rng, int random_step,
-	unsigned char *buf, fips_ctx_t *fipsctx_in)
+static int update_kernel_random(int random_step)
 {
-	unsigned char *p;
-	int fips = 0;
+	unsigned char buf[FIPS_RNG_BUFFER_SIZE]; /* random_step was checked to be <= FIPS_RNG_BUFFER_SIZE */
 	int rc;
+	struct rng *iter;
 
-	if (!arguments->ignorefail)
-		fips = fips_run_rng_test(fipsctx_in, buf);
-	if (fips && !arguments->ignorefail)
-		return 1;
+	message(LOG_DAEMON|LOG_DEBUG, "entropy successfully gathered, preparing it for the kernel\n");
 
-	for (p = buf; p + random_step <= &buf[FIPS_RNG_BUFFER_SIZE];
-		 p += random_step) {
+	while(true) {
 		if (!server_running)
 			return 0;
 		if (do_reseed) {
 			do_reseed = false;
 			alarm(arguments->force_reseed);
 		}
-		rc = random_add_entropy(p, random_step);
-		if (rc == -1)
-			return 1;
+
+		/* mix the sources on byte-level: ensure we always feed data from all available sources to the kernel
+		 * helps to mitigate problems should a source not be as random as expected */
+		int p = 0;
+		while(p < random_step) {
+			int progress = p;
+			for (int i = 0; i < ENT_MAX; ++i) {
+				iter = &entropy_sources[i];
+				if (!iter->entropy_buf.valid)
+					continue;
+
+				buf[p++] = iter->entropy_buf.entropy[iter->entropy_buf.used_pos++];
+
+				if (iter->entropy_buf.used_pos == FIPS_RNG_BUFFER_SIZE)
+					iter->entropy_buf.valid = false;
+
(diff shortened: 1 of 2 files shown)


rngd.c
```diff
@@ -752,46 +752,80 @@ static error_t parse_opt (int key, char *arg, struct argp_state *state)
...
+			 * this wastes a few bytes when FIPS_RNG_BUFFER_SIZE is not a multiple of random_step
+			 * but it makes the logic easier to implement and read */
+			if (p == progress)
+				return 0;
+		}
+
+		rc = random_add_entropy(buf, random_step);
 		if (rc == -1)
-			return 1;
+		{
```
nhorman, line 800: style nit: brace should follow the conditional, rather than be on a line by itself please

rngd.c
```diff
@@ -752,46 +752,80 @@ static error_t parse_opt (int key, char *arg, struct argp_state *state)
...
+static int random_test_sink(int random_step)
 {
+	struct rng *iter;
+
 	if (!ent_gathered)
 		alarm(1);
-	ent_gathered += FIPS_RNG_BUFFER_SIZE;
+
+	for (int i = 0; i < ENT_MAX; ++i)
+	{
```
nhorman, line 823: same nit as above

rngd.c
```diff
@@ -752,46 +752,80 @@ static error_t parse_opt (int key, char *arg, struct argp_state *state)
...
+		 * helps to mitigate problems should a source not be as random as expected */
+		int p = 0;
+		while(p < random_step) {
+			int progress = p;
+			for (int i = 0; i < ENT_MAX; ++i) {
+				iter = &entropy_sources[i];
+				if (!iter->entropy_buf.valid)
+					continue;
+
+				buf[p++] = iter->entropy_buf.entropy[iter->entropy_buf.used_pos++];
```
nhorman, line 782: given that this function only adds up to FIPS_RNG_BUFFER_SIZE of random data to the entropy pool, having each entropy source collect FIPS_RNG_BUFFER_SIZE worth of data, implies that we're going to discard FIPS_RNG_BUFFER_SIZE/ENT_MAX bytes from each entropy source on each iteration, which seems wasteful

nhorman (changes requested):

rngd.c
g-v-egidy: ah, the braces on the next line is the style i normally use. i tried to adapt to the rng-tools style, but it seems like i overlooked some cases.

I will update the pull request.

rngd.c
g-v-egidy: I'm not sure I fully understand your comment, especially the "FIPS_RNG_BUFFER_SIZE/ENT_MAX bytes" figure.

The intention of my code was to fill buf with random_step bytes (64 bytes by default) from all valid sources. then feed buf to the kernel.

After buf is fed to the kernel. the outer `while(true)` loop will continue. so if there are enough bytes left in the entropy_sources[].entropy_buf.entropy buffers they will be used. only when no source with data available is left, `p` will not progress anymore and the function will return with 0. then new entropy is gathered.

This is indeed not 100% using the available entropy, but only in the case when FIPS_RNG_BUFFER_SIZE is not an integer-multiple of random step. So in the default case of FIPS_RNG_BUFFER_SIZE = 2500 and random_step=64 we have 39 iterations (=2496 bytes) and 4 bytes left that are wasted per source. this is 0.16% which seems to be acceptable to me.

The alternative would be to decouple the buffers the data is gathered into and mixing the data. I think this will make the code more complicated and less easy to read and review. I think with a security sensitive program like rngd it is important that the code can be easily understood, reviewed and maintained.

rngd.c
g-v-egidy: fixed in new revision

rngd.c
nhorman: my concern was that we would stop when the kerenel entropy pool was full, but I guess the kernel entropy reporting is broken now anyway, so we just keep feeding until we decide we're done.  ok

g-v-egidy:
rebased patch to resolve merge conflicts.

g-v-egidy:
thanks for reviewing & merging.

PR #204, opened 2023-12-18
add --use-slow-sources option

g-v-egidy:
The entropy sources nist, jitter and pkcs11 are considered to be slow in providing entropy. By default their data is only used after all other sources failed to provide valid entropy a 100 times over. With this option rngd always tries to gather entropy from these sources too.

Having this option is useful when you want to feed the kernel data from a mix of as many independent sources as possible. The concern of for example jitter being too slow to supply a sufficient amount of entropy data is only relevant when rngd needs to supply entropy to the kernel often and in short order. With modern kernels not signaling a low watermark anymore and thus only the known force-reseed time is used, you can ensure that jitter will be able to provide enough data at configuration time.

When comparing the default timeout of 5 seconds for the jitter source with the default force-reseed time of 5 minutes, you can see that jitter will often be able to provide enough data.

Different take on this issue after discussion in #202


g-v-egidy:
thanks for reviewing & merging

Issue #198, opened 2023-07-20
JITTER_OPT_TIMEOUT default value is too low


SuperBaobab:
Hi,

I upgraded rng-tools from v6.15 to v6.16 and now the rngd service is crashing at startup because of the new 5 seconds jitter init timeout.
The jent_read_entropy time on my system is around 4 to 6 seconds and jitter is the only source of entropy.

For details, my system is on a AWS EC2 t4g.micro instance.
I think 5 seconds is too low for many systems, is it possible to increase the default value ?

Thanks,

nhorman:
Ok, so....set the timeout option on the command line.  It's all in the man page

nhorman:
Well, you're the first that's reported it since we added the option , and I'm not inclined to make it longer just to have someone on another system say it takes too long to timeout.

g-v-egidy:
While testing my changes for force_as_fast_source / --use-slow-sources I also encountered issues with the default JITTER_OPT_TIMEOUT of 5 seconds being too low **sometimes**. On two of my test systems (one being a KVM VM on a modern Epyc server, the other a lower-powered PC-Engines apu4d4 embedded board), jitter did not always initialize. It worked most of the time, but in like 15% of the cases it did time out.

I suggest to increase the time to 10 seconds.

If it is able to provide enough entropy faster, which it will on the majority of systems and program starts, then you don't have to wait longer because of this. The longer timeout will only affect systems where jitter is enabled (probably due to a default config), but always fails to initialize.

I think a 5 second longer timeout for these few systems is a good tradeoff for providing a reliable full-software fallback entropy source.

Also the jitter timeout option is currently not mentioned in the man page. I can provide a patch for one or both, once consensus reached regarding the change of the default timeout.

PR #206, opened 2023-12-19
Add jitter timeout to man page & increase jitter timeout

g-v-egidy:
Other users reported timeout issues with the current default timeout of 5 seconds in issue #198.
    
While testing jitter during development I did notice that on some systems sometimes also would fail to initialize during startup.
    
 Since these issues don't seem to be that uncommon, I think the **default timeout should be increased to 10 seconds**. Since with both reported cases the 5 second timeout was marginal, 5 seconds more should improve the probabilty of a reliable
 start on these systems by a great margin.
    
If jitter is able to provide enough entropy faster, which it will on the majority
of systems and program starts, then you don't have to wait longer because of this change.
The longer timeout will only affect systems where jitter is enabled (probably due to
a default config), but fails to initialize even after 10 seconds.

Also document the timeout option of the jitter source in the man page which was previously undocumented.

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -236,7 +236,7 @@ static struct rng_option jitter_options[] = {
 	[JITTER_OPT_TIMEOUT] = {
 		.key = "timeout",
 		.type = VAL_INT,
-		.int_val = 5,
+		.int_val = 10,
 	},
 	{
 		.key = NULL,
--- a/rngd.8.in
+++ b/rngd.8.in
@@ -237,6 +237,8 @@ Options
 
 \fBforce_soft_timer - \fR on platforms with a hardware timer that is too coarse to sample jitter, we can instead use a software based timer loop.  Detection and use of this mechanism is automatic, but this can be useful for testing purposes
 
+\fBtimeout - \fR maximum time in seconds to wait until the source must have delivered a block of entropy. On slower hardware it may be necessary to increase this value to ensure jitter can reliably be initialized (default 10)
+
 .TP
 .B
 PKCS11 (pkcs11) 


rngd.c
```diff
@@ -236,7 +236,7 @@ static struct rng_option jitter_options[] = {
 	[JITTER_OPT_TIMEOUT] = {
 		.key = "timeout",
 		.type = VAL_INT,
-		.int_val = 5,
+		.int_val = 10,
```
nhorman, line 239: I'm fine with the rest of this patch, but I'm really not comfortable with increasing this value.  We need some value, but there is absolutely no value that is going to work for everyone.  For instance, low power arm systems, or windows vms with coarse timers generally take significantly longer to generate entropy from this source.  Upping this timer just means its going to take longer for them to detect the error and move on.

nhorman (changes requested):

rngd.c
g-v-egidy: You are right that no value will work for everyone. But at least my testing and the report in #198 showed that 10 seconds would work for much more systems than the 5 seconds. And that is the idea behind a default value - be correct for the majority of use cases.

But nonetheless, I get that you really want to keep the startup time short and priorize that over losing a source.

I think jitter is the only source that is affected by this kind of problem - where the timeout during initialization is exceeded, but the source could still work fine given a longer timeout. So how about special-casing jitter here:

What I'm thinking of is to introduce a new flag for sources: `retry_initialization`. when jitter fails to initialize due to timeout, it sets this flag, but keeps the jitter threads running in the background. when the flag is set, data from the source is not used.

When there are other sources available and the `retry_initialization` flag is set, use the other sources. after the first loop through all sources is finished, retry the initialization of jitter again. the jitter threads should have had enough time in the background to get some entropy. when it still fails we abort the initialization and declare jitter dead.

When there is just jitter as source available, we use a high timeout of like 120 seconds. if not enough entropy is gathered in that time we fail and exit rngd because of no sources being available.

What do you think of this idea?

nhorman: I'm not super excited about special casing jitter like this for a few reasons:
1) Such a refactoring is going to be much larger than you think.  Letting the jitter threads run in the background is feasible, but theres still alot of initialization that occurs outside of those threads.  For example, if the AES option is selected, that happens in a blocking fashion during the init routine.  That code will have to be refactored to occur repetatively during xread calls prior to being able to use the entropy source.
2) Its going to have some wierd corner cases.  for instance, in do_loop, how do you want to determine when jitter needs to be disabled?  i.e. making a fixed time determination vs a number of iterations without entropy determiniation is going to be complex.
3) No matter what is done with jitter, eventually we need to make the timeout/keep checking decision, and thats going to largely be a per-system decision.    Trying to decide if x vs y is the better timeout value is a loosing battle.  Theres no value that I've found for which it is best for "most" systems.  There are just too many different systems with different reactions to jitter to make a qualified decision there.

That all said, I'm not opposed to having jitter work in the background, and start serving entropy when/if its ready.  If you want to write such a patch, i'll gladly review/consider it, but I would suggest the following:
1) Don't make it optional, if its going to run in the background to init, do it, lets not give users another choice for which they have to noodle over which is better for their environment.
2) make sure it honors the timeout value, lest we have it running forever chewing up cpu in the background.

g-v-egidy: >     1. Don't make it optional, if its going to run in the background to init, do it, lets not give users another choice for which they have to noodle over which is better for their environment.

ok, so jitter would always initialize late. that would indeed simplify the code.

this would have the additional benefit that rngd will most probably start faster than it does now, because even on a really fast cpu it still takes 2-3 seconds in my experience (tested on a Ryzen 7950X3D).

>     2. make sure it honors the timeout value, lest we have it running forever chewing up cpu in the background.

yes, having it fruitlessly chewing cpu forever in the background doesn't make sense.

Consider we just have jitter and no other sources available at all. So we can't just start the first gathering run without jitter and use it in the next round once it is initialized. then the jitter timeout will determine when rngd exits because no sources could be found.

Are you ok with increasing the default timeout for that case? So let's say we use something like 120 seconds for it, so that rngd tries really hard to get any source at all, but still has a reasonable timeout after which it gives up and exits.

nhorman:
i wouldn't be so sure on that in all cases, if we allow background startup and set a long timeout, there is a likelyhood that rngd will start faster, but chew cpu for a significant amount of time, leading to additional latency in subsequent operations (both inside and outside rngd).  But experiment with it, see what happens.  maybe I'm wrong

g-v-egidy:
Thanks for your ideas.

I will try implementing a deferred start for jitter in the next week or so.

PR #214, opened 2024-06-17
Add Intel CET IBT instrumentation to assembly code

nefigtut:
Add endbr64/endbr32 instruction to a function prologue. This enables IBT (Indirect Branch Tracking) feature of the Intel CET (Control-flow Enforcement Technology). All the calls and jumps in this assembly code are direct. Only functions can potentially be called indirectly. So adjusting a function prologue is enough. Add a section to indicate that this code supports IBT to linkers and security analysers.


nefigtut:
please disregard

PR #215, opened 2024-06-17
Add Intel CET IBT instrumentation to assembly code

nefigtut:
Add endbr64/endbr32 instruction to a function prologue. This enables IBT (Indirect Branch Tracking) feature of the Intel CET (Control-flow Enforcement Technology). All the calls and jumps in this assembly code are direct. Only functions can potentially be called indirectly. So adjusting a function prologue is enough. Add a section to indicate that this code supports IBT to linkers and security analysers.

### DIFF
--- a/rdrand_asm.S
+++ b/rdrand_asm.S
@@ -19,10 +19,27 @@
  *
  */
 
-#define ENTRY(x)	  \
-	.balign	64	; \
-	.globl	x	; \
-x:
+/* Add Intel CET IBT instrumentation */
+#if defined __CET__ && (__CET__ & 1)
+#ifdef __x86_64__
+	#define ENTRY(x)	  \
+		.balign	64	; \
+		.globl	x	; \
+	x:			  \
+		endbr64
+#elif defined(__i386__)
+	#define ENTRY(x)	  \
+		.balign	64	; \
+		.globl	x	; \
+	x:			  \
+		endbr32
+#endif /* __x86_64__ */
+#else /* __CET__ */
+	#define ENTRY(x)	  \
+		.balign	64	; \
+		.globl	x	; \
+	x:
+#endif /* __CET__ */
 
 #define ENDPROC(x)		  \
 	.size	x, .-x		; \
@@ -375,7 +392,35 @@ aes_round_keys:
 	.size	aes_round_keys, .-aes_round_keys
 
 /*
+ * This is necessary to inform a linker that this code has IBT (Indirect
+ * Branch Tracking) feature of the Intel CET (Control-flow Enforcement
+ * Technology) enabled.
+ * See: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection
+ */
+#if defined __CET__ && (__CET__ & 1)
+	.section	.note.gnu.property,"a"
+	.align	8
+	.long	1f - 0f		/* name length */
+	.long	4f - 1f		/* data length */
+	/* NT_GNU_PROPERTY_TYPE_0 */
+	.long	5		/* note type */
+0:
+	.string	"GNU"		/* vendor name */
+1:
+	.align	8
+	/* GNU_PROPERTY_X86_FEATURE_1_AND */
+	.long	0xc0000002	/* pr_type */
+	.long	3f - 2f		/* pr_datasz */
+2:
+	/* GNU_PROPERTY_X86_FEATURE_1_XXX */
+	.long	0x3
+3:
+	.align	8
+4:
+#endif
+
+/*
  * This is necessary to keep the whole executable
  * from needing a writable stack.
  */
-                .section        .note.GNU-stack,"",%progbits
+	.section	.note.GNU-stack,"",%progbits


rdrand_asm.S
```diff
@@ -23,6 +23,14 @@
...
 	.globl	x	; \
 x:
+/* Add Intel CET IBT instrumentation */
+#if defined __CET__ && (__CET__ & 1)
+#ifdef __x86_64__
+	endbr64
+#elif defined(__i386__)
+	endbr32
+#endif
+#endif
```
nhorman, line 33: I see this code in all the examples I can find for endbr[32|64], but Im having trouble understanding why this endbr is needed in this asm code.  My understanding is that this needs to be the next instruction run at the target of any near/far call, and the target of any indirect jmp.  But looking at the asm code I don't see any of those in the code, so I'm unsure as to why the compiler is complaining about this code.

Edit: Wait, it might be for the call in the INIT_PIC macro

rdrand_asm.S
```diff
@@ -374,8 +382,36 @@ aes_round_keys:
...
+/*
+ * This is necessary to inform a linker that this code has IBT (Indirect
+ * Branch Tracking) feature of the Intel CET (Control-flow Enforcement
+ * Technology) enabled.
+ * See: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection
+ */
+#if defined __CET__ && (__CET__ & 1)
+	.section	.note.gnu.property,"a"
+	.align	8
+	.long	1f - 0f		/* name length */
```
nhorman, line 394: Unsure I understand the magic numbers here.  why not 1 rather than 1f - 0f?  Also what name is this length referring to?

This pattern repeats below, it also seems odd to me that things like the pr_datasz is a short value.

rdrand_asm.S
```diff
@@ -374,8 +382,36 @@ aes_round_keys:
...
+	.align	8
+4:
+#endif
+
 /*
  * This is necessary to keep the whole executable
  * from needing a writable stack.
  */
-                .section        .note.GNU-stack,"",%progbits
+	.section	.note.GNU-stack,"",%progbits
```
nhorman, line 417: It seems reasonable to do this, but should it be codified in a gcc flag, so that we don't need explicit assembly for it (not sure if -mshstk is the flag needed, or if such a flag actually exists)

nhorman (commented):
Generally looks good, just a few questions to help me understand

rdrand_asm.S
nefigtut: A compiler is not complaining, it is an object verifier which reports IBT is not enabled in rdrand_asm.S and thus in the resulting rngd binary. I'm adding a section below (see below) to tell a verifier that rdrand_asm.S code supports IBT. But except _stating_ that IBT is supported we should actually _implement_ this support in code by adding ENDBR.

Yes, ENDBR should be the next instruction run after an _indirect_ near/far call and the target of any _indirect_ jmp. All the calls and jumps in this code are indeed direct. Calls to x86_rdrand_bytes() and others from C code are direct too as of now (I guess).

But in some future C code can be reworked to make an indirect call to these assembly functions. Or a compiler may decide to make a call from C code to x86_rdrand_bytes()/etc an indirect call for some reason (yes, this is unlikely). In this case and if IBT is active (it should be supported by OS and CPU and code) rngd process will be aborted without ENDBR being the first instruction of these assembly functions. So I'm adding ENDBR to the function prologue.

What puzzles me is the call to __x86.get_pc_thunk.bx. It is called directly and is local. But it is defined in a section marked both *.globl* and *.hidden*. I'm not sure what does this mean. Per my understanding .hidden means __x86.get_pc_thunk.bx cannot be seen and called outside of rdrand_asm.o object directly or indirectly. So we do not need ENDBR to __x86.get_pc_thunk.bx, as it is called only directly in rdrand_asm.S. I can be wrong here, so I'm asking for a review.

nhorman: yeah, this code was from before I took over the project.  From everything I can tell, this all looks reasonable.

Any thoughts on the 1f - 0f stuff below?

rdrand_asm.S
nefigtut: Yes, these are magic numbers, I've copied this section from https://sourceware.org/annobin/annobin.html/Test-cf-protection.html and https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/cet.h.

"not 1 rather than 1f - 0f" - this is not a number 1, this is a distance between labels 0f and 1f, i.e. a size of `.string "GNU"`. And this is a name length which is being referred to.

Same way, pr_datasz is `.long` and it is a distance between labels 3f and 2f.

It looks like the most important fields in this section is `.long   0xc0000002` (CET flags type I guess) and `.long   0x3` ((CET flags value). It is a value stating IBT and SS are supported. This section is shown as following by readelf, please note the "IBT" mark:

```
$ readelf -n rdrand_asm.o
Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000010	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 feature: IBT, SHSTK
```

rdrand_asm.S
nefigtut: I have not changed this line in any way, it was there before. I've just turned spaces into tabs to have a proper formatting.

nefigtut:
One more thing, it appears #ifdef cannot be used inside #define. Let me rework and try to force-push to update the pr.

nefigtut:
Ok, it looks like an assembly produced has proper ENDBR in the beginning of functions now. One can double-check by looking at the produced code after a pre-processor:

```
gcc -m64 -fcf-protection -fpic -c -o rdrand_asm.o -S rdrand_asm.S # x86_64 code with IBT and endbr64
gcc -m32 -fcf-protection -fpic -c -o rdrand_asm.o -S rdrand_asm.S # i386 code with IBT, endbr32 and PIC
gcc -m64 -c -o rdrand_asm.o -S rdrand_asm.S  # x86_64 code without IBT
gcc -m32 -c -o rdrand_asm.o -S rdrand_asm.S # # i386 code without IBT and PIC
```

rdrand_asm.S
nhorman: ah, their labels, not constants, thats the bit I was missing. thank you!

nefigtut:
Thanks Neil, most appreciated!

PR #224, opened 2025-12-22
Add Radiacode radiation detector as entropy source

### LINKED ISSUE #222: Question about implementation of hardware rng sources
Hi,

I'm not sure if this is the right place to ask, but is there any documentation on implementing hardware support for RNG sources and integrating it with rng-tools?

I've acquired a radiation detector/spectrometer called Radiacode with USB/Bluetooth support, and I'm curious if it would be possible to use background radiation detection information from Radiacode as a source of entropy.

I do understand that this repo is not about Radiacode and it is not what I'm asking, I will take a look at its API and documentation, I'm more interested if rng-tools would be the right place to integrate such a device/functionality and if something similar was implemented for other hardware that I could use as reference.

Is there any documentation or source code that I could use as reference? Would appreciate any information provided!

I really appreciate any help you can provide.

caquino:
# Add Radiacode Radiation Detector as Entropy Source

Closes #222

## Summary

Adds support for Radiacode radiation detectors as a hardware entropy source. Like the existing rtlsdr implementation (RF noise), this uses radioactive decay events as a physical source of randomness.

The source code for this entropy source is larger than that of the other sources, primarily due to the libusb requirement for interfacing with the radiacode device. As this is a more complex source than others, it's understandable if the maintainer chooses not to merge this PR. In the end, you will need to take care of it in the future if I'm not available to help for any reason.

Also, sorry for taking so long between opening the issue and submitting this PR; life got in the way, and libusb added a bit of complexity that required more research.

If desired, I can add more extensive documentation/notes on tests/security/code quality/reasoning.

Tested with Radiacode-102 (https://radiacode.com/)  USB: VID 0x0483, PID 0xF123, bulk transfers on 0x01/0x81

Based on the existing Python SDK, it should work with any Radiacode device.

Implementation based on:
- https://github.com/cdump/radiacode (Python SDK)
- https://github.com/kismetwireless/kismet/tree/master/capture_radiacode (C reference)

## Motivation

This adds another hardware entropy option following NIST SP 800-90B recommendations for multiple independent entropy sources.

Current sources: CPU instructions (RDRAND/RDSEED), timing jitter, rtlsdr, TPM, network-based.

| Source | Physical Basis | Hardware | USB Library |
|--------|---------------|----------|-------------|
| rtlsdr | RF noise | RTL-SDR dongle | librtlsdr |
| radiacode | Nuclear decay | Radiacode detector | libusb-1.0 |
| rdrand | Electronic noise | CPU | N/A |
| jitter | Timing variations | CPU | N/A |

Similar to rtlsdr:
- External USB hardware
- Device-specific protocol
- Vendor diversification
- Proprietary firmware

Differences:
- Decay timing vs RF noise
- Background radiation vs antenna required
- Can verify with external detectors

This is for defense-in-depth with other sources, not a replacement. 

Firmware is proprietary, like rtlsdr and CPU microcode, so it does not introduce any unknown attack vector.

## Implementation

New files:
- `rngd_radiacode.c` (The most complex part of the implementation)
- `tests/rngtestradiacode.sh`

Modified:
- `configure.ac`, `Makefile.am` - build system
- `rngd.c`, `rngd.h`, `rngd_entsource.h` - source integration
- `rngd.8.in` - documentation
- `tests/Makefile.am` - test suite

Features:
- Direct libusb-1.0 API (required as radiacode does not offer SDK/libraries that I'm aware)
- Device selection by index or serial number
- AES-128 conditioning (reuses existing ossl_aes)
- SHA-256 hashing
- Spectrum data enhancement option (320 channels)

## Testing

I added a shell script similar to the rngtestjitter.sh, but as probably not everyone will have a device available, I removed it from running by default on the Makefile.

All 4 tests pass:
```
PASS: rngtestzero.sh
PASS: rngtesturandom.sh  
PASS: rngtestjitter.sh
PASS: rngtestradiacode.sh
```

The radiacode test validates FIPS 140-2 compliance:
```bash
rngd -f -n radiacode -x rdrand -x jitter | rngtest -c 100 --pipe
```

Extended validation (1.7 GB per source):

| Blocks | Data | Kernel | Radiacode | RDRAND |
|--------|------|--------|-----------|---------|
| 1K | 2.4 MB | 2 (0.20%) | 0 (0.00%) | 0 (0.00%) |
| 10K | 24 MB | 5 (0.05%) | 4 (0.04%) | 7 (0.07%) |
| 100K | 238 MB | 85 (0.09%) | 77 (0.08%) | 89 (0.09%) |
| 500K | 1.2 GB | 402 (0.08%) | 379 (0.08%) | 405 (0.08%) |

All sources converge to ~0.08% failure rate (well within FIPS threshold). Statistical analysis shows no meaningful difference between sources (Cohen's d = 0.13).

Security validation:
- **cppcheck**: 0 issues
- **valgrind**: 0 leaks (12,593 allocs, 12,588 frees)
- **AFL++**: 449,365 executions, 0 crashes, 0 hangs
- Input validation: 21 attack vectors tested (buffer overflows, integer overflows, format strings, etc.)

## Usage

```bash
# Basic usage (device 0)
sudo rngd -n radiacode

# Select by index
sudo rngd -n radiacode -O radiacode:device_id:1

# Select by serial
sudo rngd -n radiacode -O radiacode:serial:RC-102-000123

# Adjust poll delay (ms)
sudo rngd -n radiacode -O radiacode:poll_delay:200

# Combine with other sources
sudo rngd -n rdrand -n radiacode
```

Options (format: `-O source:key:value`):
- `device_id` (0-99, default 0): Device index
- `serial` (string): Serial number, overrides device_id
- `poll_delay` (1-10000, default 100): Milliseconds between polls
- `use_spectrum` (0-1, default 1): Enable spectrum enhancement

Multiple devices are listed at startup:
```
Radiacode devices found:
0: VID:0483 PID:f123 Serial:RC-102-000123
1: VID:0483 PID:f123 Serial:RC-102-000124
Using device 0
```

Note: One device per rngd instance.

## Requirements

Build: libusb-1.0-dev (indirectly required for rtlsdr, so not exactly a new dependency), OpenSSL (already needed)

Runtime: Radiacode device, Linux with `/dev/random`, USB permissions

```bash
./autogen.sh
./configure --prefix=/usr --sysconfdir=/etc  # or --without-radiacode to disable
make
sudo make install
sudo make check
```

## Notes

Performance: ~256 bits per read, <10ms USB latency

Environment: Works with background radiation (0.1-0.2 μSv/h typical indoors). Shielding reduces the rate. 
Optional enhancements: granite nearby, smoke detectors (Am-241). Any radiation source near the device increases data generation.
Testing used only background radiation.

Compatibility: No changes to existing sources. Opt-in with `-n radiacode`. Builds without a device present. Graceful fallback if the device is missing.

## Code Quality

Static/dynamic analysis clean:
- cppcheck: 0 issues
- valgrind: 0 leaks, 0 errors
- AFL++: 449K execs, 0 crashes
- No unsafe functions (strcpy, sprintf, etc.)
- All allocations checked
- Bounds checking on all access
- Integer overflow protection

This code tries to follow the same pattern used on other sources, especially the rtlsdr one.

USB disconnection handling: Like rtlsdr, returns error on `libusb_bulk_transfer()` failure. rngd doesn't crash - continues running, retries automatically, recovers when reconnected. You don't need to do any manual restart.

### DIFF
--- a/rngd.c
+++ b/rngd.c
@@ -171,6 +171,7 @@ static enum {
 	ENT_RTLSDR,
 	ENT_QRYPT,
 	ENT_NAMEDPIPE,
+	ENT_RADIACODE,
 	ENT_MAX
 } entropy_indexes __attribute__((used));
 
@@ -337,6 +338,32 @@ static struct rng_option namedpipe_options[] = {
 	}
 };
 
+static struct rng_option radiacode_options[] = {
+	[RADIACODE_OPT_DEVID] = {
+		.key = "device_id",
+		.type = VAL_INT,
+		.int_val = 0,
+	},
+	[RADIACODE_OPT_POLL_DELAY] = {
+		.key = "poll_delay",
+		.type = VAL_INT,
+		.int_val = 100, /* 100ms default poll delay */
+	},
+	[RADIACODE_OPT_USE_SPECTRUM] = {
+		.key = "use_spectrum",
+		.type = VAL_INT,
+		.int_val = 1, /* Enable spectrum entropy by default */
+	},
+	[RADIACODE_OPT_SERIAL] = {
+		.key = "serial",
+		.type = VAL_STRING,
+		.str_val = "", /* Empty string means use device_id index */
+	},
+	{
+		.key = NULL,
+	}
+};
+
 static struct rng entropy_sources[ENT_MAX] = {
 	/* Note, the special char dev must be the first entry */
 	{
@@ -484,6 +511,22 @@ static struct rng entropy_sources[ENT_MAX] = {
 		.xread	  = xread_namedpipe,
 		.init	   = init_namedpipe_entropy_source,
 		.rng_options	= namedpipe_options,
+	},
+	{
+		.rng_name	= "Radiacode radiation detector entropy source",
+		.rng_sname	= "radiacode",
+		.rng_fd		= -1,
+		.flags		= {
+			.slow_source = 1,
+		},
+#ifdef HAVE_RADIACODE
+		.xread		= xread_radiacode,
+		.init		= init_radiacode_entropy_source,
+		.close		= close_radiacode_entropy_source,
+#else
+		.disabled	= true,
+#endif
+		.rng_options	= radiacode_options,
 	}
 
 };
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,11 +32,15 @@ if RTLSDR
 rngd_SOURCES	+= rngd_rtlsdr.c
 endif
 
+if RADIACODE
+rngd_SOURCES	+= rngd_radiacode.c
+endif
+
 if QRYPT
 rngd_SOURCES	+= rngd_qrypt.c
 endif
 
-rngd_LDADD	= librngd.a $(LIBS) $(librtlsdr_LIBS) ${libp11_LIBS} ${libcrypto_LIBS} ${jansson_LIBS} ${libcurl_LIBS} ${libxml2_LIBS} ${openssl_LIBS} ${libcap_LIBS} $(PTHREAD_LIBS)
+rngd_LDADD	= librngd.a $(LIBS) $(librtlsdr_LIBS) $(libusb_LIBS) ${libp11_LIBS} ${libcrypto_LIBS} ${jansson_LIBS} ${libcurl_LIBS} ${libxml2_LIBS} ${openssl_LIBS} ${libcap_LIBS} $(PTHREAD_LIBS)
 
 if DARN
 rngd_SOURCES	+= rngd_darn.c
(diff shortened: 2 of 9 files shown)


caquino:
As I assume you will not have any device available to test, I'm more than happy to run any necessary tests.

caquino:
@abajk I tried to do a quick code here that keeps trying to read the device continuously. The device always returns a timestamp and sequence number, so if multiple reads return the same sequence number, I discard that read because it didn't generate new data.

These are the results:
```
Radiacode speed test (duration: 60 seconds):
  Total bytes: 3316
  Requests: 66076
  Min bytes/request: 0
  Max bytes/request: 714
  Avg bytes/request: 0.05
  Throughput: 0.00 Mbps
```
Actually doing this test made me realise that I also need to take this into account on the entropy source, as the sequence number and timestamps are not to be considered random data, as they are predictable.

I just realised that since it does not generate much data from background radiation, I can use the interval between two different sequence numbers (or radiation-decay events) to add extra data to the RNG.

Also, the radiacode has what is called "spectrum data", and I will also try to gather it to use to increase the data it generates. I was using the event data, which has the last event, the spectrum data are like buckets with energy distribution of the events, as the device is more than a radiation detector, it is a scintillator. 

The current PR reads only the last events since the previous read. By incorporating spectrum information, it will reveal the radiation "signature" of the source. Basically, which elements are being detected.

This will not increase the source's throughput, but it will increase the randomness it provides.

I will work on the code changes and update the PR based on the findings, but given this speed, I assume the marked slow source is correct. Is my understanding correct?

abajk:
That sounds reasonable, but it would be good to hear other people's opinions as well.

caquino:
Another point: I just realised we were taking a snapshot of the last available data at the moment it was requested from the entropy source, but any events that happened in between reads were not used at all. After reading the `rngd_jitter.c`, it looks like it uses a thread to read and "accumulate" the data. This would be a better option for this source; also, this way, we can accumulate events and have more random data. I will work on this over the weekend and update this PR.

rngd.c
```diff
@@ -484,6 +511,22 @@ static struct rng entropy_sources[ENT_MAX] = {
 		.xread	  = xread_namedpipe,
 		.init	   = init_namedpipe_entropy_source,
 		.rng_options	= namedpipe_options,
+	},
+	{
+		.rng_name	= "Radiacode radiation detector entropy source",
+		.rng_sname	= "radiacode",
+		.rng_fd		= -1,
+		.flags		= {
+			.slow_source = 1,
```
nhorman, line 520: given that you are conditioning the buffer with an aes encryption round, does this really need to be a slow source?  It seems like it can provide entropy at whatever rate you like.

nhorman (changes requested):
Generally this looks pretty good, thank you for doing it.  just the one question regarding the slow source usage

nhorman:
meh, the one issue I have isn't really critical, I can take this as is

